PR #2, opened 2018-06-14
Get Travis build working

msc-:


msc-:
This is no good. The push to docs triggers another build.

PR #19, opened 2018-11-16
 Refine levelImages layout

### LINKED ISSUE #16: Tighten levelImages layout
`levelImages` structure is:
```
for each array_element in numberOfArrayElements 
   for each face in numberOfFaces 
       for each z_slice in pixelDepth 
           for each row or row_of_blocks in pixelHeight 
               for each pixel or block_of_pixels in pixelWidth
                   Byte data[format-specific-number-of-bytes] 
               end
           end
       end
   end
end
```

Since `format-specific-number-of-bytes` may be unknown, an additional restriction is needed:
`bytesOfUncompressedLevelImages % (numberOfArrayElements * numberOfFaces) == 0`

Also `z_slice` should be changed to something like `z_slice or rows_of_blocks` to accommodate compressed 3D formats.

lexaknyazev:
Fixes #16.

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -875,8 +875,8 @@ Aaftab Munshi. The Khronos Group, July 2003.
   OpenGL^®^ Graphics System, A Specification (Version 4.6 (Core Profile))].
 Mark Segal, Kurt Akeley; Editor: Jon Leech. The Khronos Group, July 2017.
 
-- [[[VULKAN11]]] https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html[Vulkan^®^
-1.1.87 Specification].
+- [[[VULKAN11]]] https://www.khronos.org/registry/vulkan/specs/1.1/html/vkspec.html[Vulkan^®^
+1.1 Specification].
 The Khronos Group, October 2018.
 
 ////
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -11,6 +11,7 @@
 :docinfo1:
 :doctype: article
 :encoding: utf-8
+:stem:
 // Disabling toc and numbered attributes doesn't work with a2x.
 // Use the xsltproc options instead.
 :toc!:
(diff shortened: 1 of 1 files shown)


MarkCallow:
Thank you for this. I did not like several aspects of the layout: the equations and definitions being too much for a callout and having all 3 equations for num_blocks on a single line. I would also like the variables in the equations to be rendered in italic. After working on this for a while here is what I've come up with.

screen shot 2018-11-17 at 17 44 29

Unfortunately I could not figure out how to get the variables rendered in italic with AsciiMath, nor could I figure out how to indent the num_blocks equations so I rewrote them in latexmath which makes the Asciidoctor source of the equations pretty much unreadable. If you can figure out how make something like the above using AsciiMathML please do it. Here is my code for the above. The documentation for AsciiMath is remarkably, er, minimal not that the documentation for latexmath is much better.

```
<1> Replace with 1 if this field is 0.
<2> Must be 1 if `glInternalFormat` is one of the `GL_PALETTE*` formats
    from GL_OES_compressed_paletted_texture <<OESCPT>>.
<3> See <<levelImages_defs,the definitions>> below.
<4> Rows of uncompressed texture images must be tightly packed,
    equivalent to a `GL_UNPACK_ALIGNMENT` of 1.

[[levelImage_defs]]In the `levelImages` loops above,

[stem]
+++++
num\_blocks\_z = \max\left(1, \left\lceil{\frac{pixelDepth}{block\_depth}}\right\rceil\right)
+++++

[stem]
+++++
num\_blocks\_y = \max\left(1, \left\lceil{\frac{pixelHeight}{block\_height}}\right\rceil\right)
+++++

[stem]
+++++
num\_blocks\_x = \left\lceil{\frac{pixelWidth}{block\_width}}\right\rceil
+++++

where _block_depth_, _block_height_, and _block_width_ are `1` for
uncompressed formats and the block size in that dimension for block
compressed formats as given in the format's section of the Khronos
Data Format specification <<KDF12>>.

A _block_ is a single pixel for uncompressed formats and
stem:[block\_width \times block\_height \times block\_depth]
pixels for block compressed formats.
```

Note, it is necessary to change `:stem:` in the doc header to `:stem: latexmath` for this to work.

lexaknyazev:
> I could not figure out how to get the variables rendered in italic with AsciiMath

AFAIU, AsciiMath uses italic font for one-letter variables and regular font for "multi-letter text". There are six font commands but italic is not among them.

Unless we want to rewrite these equations like 
image
we have to switch to LaTeX.

Please let me know your preference.

MarkCallow:
We use latexmath and put readable equations in comments where necessary.

I will my include my changes when I merge this PR.

PR #23, opened 2018-11-22
Disallow BOM for metadata keys

lexaknyazev:
For a better compatibility, UTF-8 strings must not have BOM. 
This guarantees that, e.g., `KTXswizzle` key will always be written as
```
4B 54 58 73 77 69 7A 7A 6C 65 00
```
otherwise, implementations would need to support BOM-version as well:
```
EF BB BF 4B 54 58 73 77 69 7A 7A 6C 65 00
```

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -498,16 +498,16 @@ the bytes in `<<valuePadding>>`.
 
 === keyAndValue
 `keyAndValue` contains 2 separate sections. First it contains a key
-encoded in UTF-8 without BOM. The key must be terminated by a NUL
-character (a single 0x00 byte). Keys that begin with the 3 ascii 
-characters 'KTX' or 'ktx' are reserved and must not be used except 
-as described by this spec (this version of the KTX spec defines two 
-keys). Immediately following the NUL character that terminates the key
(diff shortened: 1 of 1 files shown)


ktxspec.adoc
```diff
@@ -498,18 +498,18 @@ the bytes in `<<valuePadding>>`.
 
 === keyAndValue
 `keyAndValue` contains 2 separate sections. First it contains a key
-encoded in UTF-8. The key must be terminated by a NUL character (a
-single 0x00 byte). Keys that begin with the 3 ascii characters 'KTX'
-or 'ktx' are reserved and must not be used except as described by
-this spec (this version of the KTX spec defines two keys).
-Immediately following the NUL character that terminates the key is
-the Value data.
+encoded in UTF-8 without BOM. The key must be terminated by a NUL
```
MarkCallow:
```suggestion
encoded in UTF-8 without a BOM (Byte Order Mark). The key must be terminated by a NUL
```

MarkCallow:
Thanks. It is good to clarify this even though the Unicode spec states that a BOM has no utility with UTF-8 and recommends a BOM not be used.

PR #26, opened 2018-11-24
Spec has wrong ascii values for KTX22

Karanlos:
Changed byte array identifier hex values to match what the spec doc says the array should contain.

Question: Should ktx-mime-registration.txt also be updated to reflec the new value?


MarkCallow:
Thanks for spotting this.

Issue #14, opened 2018-10-31
How to handle type-awareness disparity between DFD and `VkFormat`


MarkCallow:
The Data Format Descriptor has no awareness of primitive types. A DFD describes the data as it is laid out in memory and is thus independent of endianness. However Vulkan & GL formats are type aware and, when the file writer and reader have different endianness, those with primitive type sizes of 2, 4 or 8 bytes must be byte-swapped into the reader's endianness before the data can be uploaded to GL or Vulkan using the vkFormat specified in the KTX file.

Note that there are a few VK_FORMATS which have opposite-endian equivalents that would enable uploads to be done without any byte swapping, e.g `VK_FORMAT_R8G8B8A8_UNORM` and `VK_FORMAT_A8B8G8R8_UNORM_PACK32`. These cases are very much the minority so it doesn't change the need to resolve this issue.

Changing the endianness of DFD-described data requires

- recognizing that the data has a size that requires swapping which is somewhat involved.
- modifying the DFD to describe the changed-endian data
- byte-swapping the data.

Andrew G. has promised some sample code to do this.

The DFD spec. is supporting a lot wider range of data layouts than KTX2 needs which means Andrew finds making DFD type aware unattractive.  However placing restrictions on the DFDs used in KTX files is theoretically possible. I'm not sure what restrictions we could make in practice that would help here.

Meanwhile the KTX2 header is also type-aware. It has 32- and 64-bit fields so we can't remove the `endianness` field and swapping requirement we already have.

How to we handle the disparity?

1. Require that the image data should always be little-endian. Drawbacks here are that big-endian systems would have to swap bytes on both writing and reading KTX files and writers on big-endian machines are very likely to overlook this requirement leading to invalid files.
2. When the endianness field indicates a mismatch require readers to determine from the DFD if the data needs swapping, byte swap it and rewrite the DFD.
3. Retain a `typesize` field which is used when `vkFormat != VK_UNKNOWN_FORMAT`.  In this case the DFD must be rewritten as well. When it is VK_UNKNOWN_FORMAT` let the reader decide how it wants to handle the data.
4. Some other way?

Subject to seeing the code sample from Andrew, I am leaning towards 2.

jherico:
I'm a little confused...

> Retain a typesize field which is used when vkFormat != VK_UNKNOWN_FORMAT.

[...]

Issue #37, opened 2018-12-23
Better ASTC support


lexaknyazev:
ASTC texture formats cover lots of use-cases and KTX2 spec requires additional details to fully support them.

`vkFormat` (as well as `glFromat`) gives only three properties of ASTC data: 
 * Block dimension (2D or 3D)
 * Block footprint size (4x4, 5x4, ...)
 * sRGB output enabled or not

**Blocking issue:**
- There's no Vulkan enums for ASTC 3D blocks. 
  - [ ] @dewilkinson or @MarkCallow, could you ask Vulkan WG about this?

**Not every format value is valid for every possible KTX2 dimensions configuration.** Moreover, LDR/HDR modes are signaled  _per-block_ within texture data.

- ASTC is not supported for 1D textures.
  - [x] This is already enforced by KTX2 dimensions restrictions.
- There are additional constraints based on block sizes - i.e., `pixelWidth` must be a multiple of block's width.
  - [x] We can enhance `pixelWidth` and friends section with that because it's relevant for all block formats.
- Implementations supporting only LDR profile (without `_sliced_3d` extension) will not accept ASTC for 3D texture targets.
  - [ ] Not a file format issue, but maybe it's worth mentioning.
- sRGB formats cannot contain blocks with HDR modes (decoder will output error color).
  - [x] We can use DFD to tighten this. See (section 7.5):
    >  ... an ASTC texture that is guaranteed by the user to contain only LDR-encoded blocks should have the `channel_id` `KHR_DF_SAMPLE_DATATYPE_FLOAT` bit clear, ...
- Some recent mobile GPUs support reducing ASTC's decoder precision for better cache performance and power saving. While the encoded data remains the same, the exact decoded bits will be different, so some pipeline tools may want to keep this information.
  - [x] If there's enough interest, we can add a new metadata entry (see `VK_EXT_astc_decode_mode` or `EXT_texture_compression_astc_decode_mode`).

MarkCallow:
> It appears to be the same in Vulkan

It may not be. I'm checking...

lexaknyazev:
>I have not been able to find the `sliced_3d` extension. Is this a GL extension?

Yes, it's called `KHR_texture_compression_astc_sliced_3d`. I didn't manage to find any info re Vulkan support.

> `VK_EXT_astc_decode_mode` looks to me like the way to enable true LDR...

[...]

Issue #8, opened 2018-10-25
Tighten Vulkan / GL enums


lexaknyazev:
The KTX2 headers includes 3 OpenGL-originated fields describing data format (`glType`, `glFormat`, and `glInternalFormat`) and one from Vulkan (`vkFormat`). The presence of all four of them drastically increases the amount of possible combinations and may lead to various implementation issues unless strictly defined by the KTX2 spec.

Here're the most outstanding issues, as I see them.

### `vkFormat` alone is usually enough to fully describe data
For example: `VK_FORMAT_R8G8_UINT`
- `glType`: UNSIGNED_BYTE
- `glFormat`: GL_RG_INTEGER
- `glInternalFormat`: GL_RG8UI

It's not clear what implementations should do if they encounter mismatching GL values with known Vulkan format.

### Desktop OpenGL allows in-driver data conversion on texture upload
So `glInternalFormat` don't have to match other fields to be consumed by OpenGL API. Behavior may vary by vendor and driver version.

### Vulkan has a notion of "scaled" formats which has no direct equivalent in OpenGL ES
For example: `VK_FORMAT_R8G8_USCALED`. There's no valid `glFormat` / `glInternalFormat` combination to cover this. Since such formats are not mandatory in Vulkan, they could be disallowed in KTX2.

### Vulkan supports sRGB with RED and RG formats
See `VK_FORMAT_R8_SRGB` and `VK_FORMAT_R8G8_SRGB`.
There's no specific `glInternalFormat` enums for them in the core OpenGL. Desktop OpenGL may be able to accept `GL_SRGB8` if conversion works. For embedded, there're `EXT_texture_sRGB_R8` and `EXT_texture_sRGB_RG8` extensions. Which of these should KTX2 allow?

### BGRA format requires different enums on OpenGL and OpenGL ES
Vulkan has all flavors of it: signed/unsigned/integer/normalized/sRGB. OpenGL supports `GL_BGRA` for `glFormat` but there's no corresponding internal format. On the other hand, OpenGL ES may accept `GL_BGRA8_EXT` as `glInternalFormat` with `APPLE_texture_format_BGRA8888` extension. BGR without alpha has even worse API support.

### Some compressed formats from OpenGL don't have matching `vkFormat`.
- `OES_compressed_paletted_texture`
- `AMD_compressed_3DC_texture`
- `AMD_compressed_ATC_texture`
- `3DFX_texture_compression_FXT1`
[...]

PR #40, opened 2019-01-08
Issue8: Remove GL format info.

MarkCallow:
Please review. Note the format mapping table still needs to be filled in.

Before we can finalize this we need to resolve aspects of issue #37, specifically get some confidence that Vulkan extensions for ASTC HDR and 3D will emerge.

We need to find a reasonable message for handling of future formats.

In KTX, because we used GL enums and because GL does not require extensions to be enabled, an application/loader that just passed format, internalformat and type to GL could load formats that did not exist when it was written. As long as it was checking for GL errors, it would fail gracefully if the implementation did not support the new format.

Since Vulkan extensions must be enabled, and physical device features enabled for future compression schemes added to core, apps will not be able to handle future formats without modification. There will also not be a mapping to other APIs in the table in the spec.

Making a new KTX version when (if) new formats emerge is unacceptable. The mapping table does not affect the data in the file so it is okay to update the table with new formats from time to time. Existing applications may not be able to load these in Vulkan, as they won’t be enabling the feature/extension or in other APIs because they won’t have the mapping. Some questions:

1. Should we put the mapping in metadata? Basically move the gl* values to metadata and possibly add mappings to other API as well?

2. Should we just accept that older applications won’t be able to load newer formats?

In PR at present I have written

```
The table in <<formatMapping>>
gives the mapping for all `VkFormat` enum values, known at the time
of writing, to OpenGL internal format, format and type values.
Applications must use this mapping.
```

How do we allow for the possibility that there may not be a mapping on the table? This becomes moot if we put the mapping in metadata.

We considered putting the needed extension or feature name in metadata and this PR contains such metadata so reviewers can see what it would be like. However since features and extensions must be enabled at `VkDevice` creation, likely long before any KTX file is read, the utility is questionable. We'd like feedback.

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -284,15 +284,15 @@ is a not a recognized Vulkan format.
 .Rationale
 ====
 The A8B8G8R8*PACK32 formats are prohibited because the end result
-is the same regardless of whether the data is treated as packed into
-32-bits or as the equivalent R8G8B8A8 format, i.e. as an array of
-4 bytes, and because the Data Format Descriptor cannot distinguish
-between these cases.
+is the same regardless of whether the data is treated as packed
+into 32-bits or as the equivalent R8G8B8A8 format, i.e. as an array
+of 4 bytes, and a Data Format Descriptor cannot distinguish between
+these cases.
 
-The \*SCALED* formats are prohibited because they are not widely
-supported for texturing and because the Data Format Descriptor
-cannot distinguish tbese from the int values having the same bit
-pattern.
+The \*SCALED* formats are prohibited because they are intended for
+vertex data, very few, if any, implementations support using them
+for texturing and a Data Format Descriptor cannot distinguish
+these from `int` values having the same bit pattern.
 ====
 
 === [[dimensions]]pixelWidth, pixelHeight, pixelDepth
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -268,10 +268,11 @@ There are not yet Vulkan extensions for the ASTC HDR and 3D formats
 described in _OES_texture_compression_ASTC_ <<OES_ASTC>>. ASTC
 formats are indicated in the DFD by setting `color_model` to
 `KHR_DF_MODEL_ASTC (= 162)`. HDR data is indicated by setting the
-`channel_id` `KHR_DF_SAMPLE_DATATYPE_FLOAT` bit to 1. An ASTC 3D
-texture is indicated by `texel_block_dimension_2` > 0. Tools handling
-ASTC and OpenGL loaders must be be able to recognize these formats
-from the DFD.
+`channel_id` `KHR_DF_SAMPLE_DATATYPE_FLOAT` bit to 1. The block
+size is given by the values of `texture_block_dimension_0` and
+`texture_block_dimension_1` and an ASTC 3D texture is indicated by
+`texel_block_dimension_2` > 0. Tools handling ASTC and OpenGL loaders
+must be be able to recognize these formats from the DFD.
 
 [TIP]
 ====
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -264,6 +264,15 @@ formats?
 is a not a recognized Vulkan format. The data layout is always given by
 the Data Format Descriptor.
 
+There are not yet Vulkan extensions for the ASTC HDR and 3D formats
+described in _OES_texture_compression_ASTC_ <<OES_ASTC>>. ASTC
+formats are indicated in the DFD by setting `color_model` to
+`KHR_DF_MODEL_ASTC (= 162)`. HDR data is indicated by setting the
+`channel_id` `KHR_DF_SAMPLE_DATATYPE_FLOAT` bit to 1. An ASTC 3D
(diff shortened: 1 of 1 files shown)


lexaknyazev:
> Should we put the mapping in metadata?

Since this introduces data redundancy, such metadata could be reliably used only in trusted environments (so that GL metadata matches VkFormat and doesn't need verification on load). This doesn't make much benefit for already existing formats because such mapping will likely be a part of a loader anyway.

> Should we just accept that older applications won’t be able to load newer formats?

Pipeline and content creation tools tend to update regularly these days, so it shouldn't be a problem. Shipped games don't get updates with texture data of unknown format, so they should be safe as well.

MarkCallow:
> Since this introduces data redundancy, such metadata could be reliably used only in trusted environments (so that GL metadata matches VkFormat and doesn't need verification on load). This doesn't make much benefit for already existing formats because such mapping will likely be a part of a loader anyway.

I do not truly understand this concern. Why would an app trust the supplied VkFormat but not the GL metadata?

> Pipeline and content creation tools tend to update regularly these days, so it shouldn't be a problem. Shipped games don't get updates with texture data of unknown format, so they should be safe as well.

Good points.

I've just added a new commit with some language on handling of ASTC HDR & 3D in the event Vulkan extensions are not released by the time we release KTX2. Please review.

MarkCallow:
This PR is for issue #8. Previous comments fails to link it to the issue.

lexaknyazev:
> Why would an app trust the supplied VkFormat but not the GL metadata?

VkFormat is a required field that _defines_ what format texture is. GL metadata is optional by design - some clients will read it, some ignore. Potential behavior fragmentation is my concern here.

FWIW, I'd propose omitting GL metadata for formats that already have GL mappings. At the same time, I'm fine with GL (or other API) metadata when VkFormat is undefined.

ktxspec.adoc
```diff
@@ -815,38 +818,47 @@ Although other orientations can be represented, it is recommended
...
+==== KTXdxgiFormat
 
-- `VK_IMG_format_pvrtc`
+For Direct3D the mapping is specified with the key
 
-Loaders are not required to enable the extension if they determine that
-the Vulkan implementation supports use of the format without it. This
-can happen, for example, if the implementation is of a later version of
-Vulkan where the format has become core.
+- `KTX__dxgiFormat`
```
lexaknyazev: Why use underscore here?

ktxspec.adoc
```diff
@@ -815,38 +818,47 @@ Although other orientations can be represented, it is recommended
...
-be enabled to use the format given by `vkFormat`, it must specified
-in metadata with the key
+==== KTXmtlPixelFormat
 
-- `KTXrequiredVkPhysicalDeviceFeature`
+For Direct3D the mapping is specified with the key
 
-The value is the name of the feature expressed as a string in the ASCII
-subset of UTF-8, e.g.,
+- `KTXmetalPixelFormat`
```
lexaknyazev: Should the key be `KTXmtlPixelFormat`?

ktxspec.adoc
```diff
@@ -815,38 +818,47 @@ Although other orientations can be represented, it is recommended
...
 
-Support for some standard formats is optional in Vulkan. Support
-must be queried via `VkGetPhysicalDeviceFeatures2` and, if supported,
-the feature enabled at `VkDevice` creation time. If a feature must
-be enabled to use the format given by `vkFormat`, it must specified
-in metadata with the key
+==== KTXmtlPixelFormat
 
-- `KTXrequiredVkPhysicalDeviceFeature`
+For Direct3D the mapping is specified with the key
```
lexaknyazev: Should it be Metal here?

MarkCallow:
I have just added 2 more commits. This PR is now complete subject to review. Please take a look at the whole change.

ktxspec.adoc
MarkCallow: To make the length a multiple of 4 so the UInt32 is aligned. There is a note in the introduction to the section.

ktxspec.adoc
MarkCallow: Again I want the name to be a multiple of 4 (including the terminating NUL).

ktxspec.adoc
MarkCallow: Yes. Fixed in a commit I just sent. Thanks for spotting it.

MarkCallow:
I plan to merge this on Monday (Jan 21st). Please speak now if you see problems.

mosra:
Just a note from a random guy passing by (sorry if this was discussed elsewhere already and I missed it):

There's the `glTypeSize` field, last one that survived the removal. Its description references the now-gone `glType` field, so maybe that needs to be updated as well? Besides that, what about naming it just `typeSize`? Since there are no other `gl`-prefixed fields anymore and the size is not specific to GL (same size would be for VK as well) and it rather describes endianness-specific handling.

lexaknyazev:
Good catch. I expected that field removed since we've agreed on little-endian only.

MarkCallow:
> There's the `glTypeSize` field, last one that survived the removal.
Thanks for the note. I plan to remove this when I fix issue #14, not as part of this issue. I didn't rename it because of the planned removal.

Issue #39, opened 2018-12-31
Relocate Supercompressed Global Data to start of bitstream block


dewilkinson:
It has been decided to remove the binary Global Data block from the current location under the header of KTX2 file, and move it to its own block at the start of the bitstream data, immediately before the highest mip level data offset. The datastream will therefore look like this:

<start texturedatablock>
Pos 0: Supercompressed Global Data [n]
Pos n:  Miplevel m [n(m)]
Pos n + n(m) : Miplevel m-1 [n(m-1)]
Pos n  + n(m) + n (m-1): .......
...
<end texturedatablock>

MarkCallow:
This is fine but I still plan to include the offset to it in the index which I thought was the understanding we had reached.

dewilkinson:
Yes, absolutely, this should have been included in the above resolution. Offset will still be required

dewilkinson:
Mark: what label do we switch to when issue is resolved, but waiting for implementation?

MarkCallow:
None. The fact that it is still open indicates awaiting implementation or fix.

MarkCallow:
I'll be fixing issue #32 along with this in a single change.

MarkCallow:
Please review PR #53.

MarkCallow:
Fixed by PR #53.

Issue #32, opened 2018-12-09
Format needs to be more directly seekable


dewilkinson:
ISV feedback:

“The format is non-seekable. Excellent work that I need to read through half the file so I can start skipping to the bits I want. Point in case, I just want the lowest level mip-map. I read the file header, then I need another request to read the bytesOfDataFormatDescriptor worth of data (and the bytesOfKeyValueData), which I can then use to skip to bytesOfSupercompressionGlobalData, then I can skip over that. Which means I need two seeks and reads, or, if I’m doing HTTP, three requests assuming I don’t download the whole file right away (assuming a file with both supercompressed and non-supercompressed data is valid.)“

MarkCallow:
Putting aside the hyperbole, I'm open to putting an index immediately after the header.

The miplevels are currently indexed. We could combine this and the new index at the front of the file. Since the number of miplevels is unknown until after the header has been read, the header must be read before the length of the index will be known. So 2 reads will be necessary to read in the complete index and a  seek and a third read to get level data. If we don't fold in the level index, you'll need a read and a seek to find the level index, a second read to get the index and another seek and a third read to get the level data. So an index doesn't seem to provide any reduction in the number of reads & seeks.

The comments about "doing HTTP" suggest that you can do seeks and partial reads via HTTP. Is this true?

jherico:
> The comments about "doing HTTP" suggest that you can do seeks and partial reads via HTTP. Is this true?

@MarkCallow yes, HTTP supports [range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests) to fetch a subset of an asset.  Our application already uses this with KTX 1 to fetch a header, and then incrementally fetch the mips

dewilkinson:
Reviewed in 12/21 call

It was agreed that an index table block after the KTX2 header is a good solution to allow a reader to seek to any specific data block in the file

MarkCallow:
I intend to fix this in the same change as issue #39 as they changes are closely related so I am closing this.

MarkCallow:
Please review PR #53.

PR #53, opened 2019-02-05
Reorder fields

MarkCallow:
This reorders the fields addressing issue #32 and issue #39. Please review.

I am aware the glTypeSize and endianness fields are still present. I'll address those next.

I have made the file offsets in the index all relative to the start of the file. I can't see any compelling reasons for choosing any particular one of the possibilities: start of file, start of index or end of index as the base. Comments please.

This is a large textual change so looking at diffs will probably not be helpful. I suggest you build the spec and read the HTML.

### DIFF
--- a/docinfo.html
+++ b/docinfo.html
@@ -15,6 +15,10 @@ div.legal p {
   font-size: 60%;
 }
 
+.optional {
+  opacity: 0.5;
+}
+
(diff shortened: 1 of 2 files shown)


MarkCallow:
I couldn't wait any more for reviews.

PR #57, opened 2019-02-19
[WIP] Cleanup remaining padding fields

lexaknyazev:
This PR clarifies the number of `mipPadding` bytes.

Also, `dfdTotalSize` section refers to `dfdBlockPadding` field, but it doesn't exist anywhere. Should it be removed from the description or added to the File Structure?

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -697,8 +697,8 @@ A _dfDescriptor_ is useful in the following cases:
 
 ==== dfdTotalSize
 Called `total_size` in <<KDF13>>, `dfdTotalSize` indicates the total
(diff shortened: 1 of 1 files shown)


MarkCallow:
dfdBlockPadding should be removed from the description. DFD 1.3 will mandate that descriptor sizes are multiples of 4 so it is no longer needed. Please update the PR.

lexaknyazev:
Uint64 fields in the header are not aligned to 8-byte boundaries. Should it be addressed?

MarkCallow:
I will fix the alignment issue. I have a couple of changes in the pipeline which affect it, which is why it is wrong at present.

PR #58, opened 2019-02-20
First draft of format mappings

### LINKED ISSUE #44: Fill out format mapping table
Format mapping needs to be filled out from [this data](https://docs.google.com/spreadsheets/d/1t6xFXyhlrLnSVucg6bc_Gh5yrDVKZlQLH897NnsTpTY/edit?usp=sharing).

lexaknyazev:
Fixes #44.

All data and supporting boilerplate is in place, we can tweak styling if needed. I chose to use list form instead of a table because it looks better on smaller screens.

JSON file contains additional information about each texture format such as type and block size. These properties are unused, but they still may be helpful. PVRTCv1 formats have block size of zero because these formats have more complex layout. Multi-plane and video formats are not included for now.

JSON-Schema is included for developers who may want to use code generation.


.gitignore
```diff
@@ -1,2 +1,3 @@
+.idea
```
MarkCallow: What is this file? If it's temporary make clean ought to remove it.

lexaknyazev: This is a folder that IDE creates locally. This line could be safely removed.

MarkCallow: What is IDE? Integrated development environment? If so which one?

lexaknyazev: Yes, it's IntelliJ IDEA. I could move this line to my global `.gitignore` to keep this repo clean.

MarkCallow:
It looks good but it is soooo long. I guess there's not much we can do about it. Given the lengths of the some of the token names any kind of table, e.g. a table under each Vulkan format heading, is going to be too wide.

lexaknyazev:
With some extra HTML/JS/CSS we can make this list collapsible, so that it's hidden by default.

MarkCallow:
Thanks for moving the .idea ignore.

Making the list collapsible would be great. Do you want to add it to this PR or some I accept this and let you do it in a different PR?

MarkCallow:
It occurs to me as I begin implementation that we'll need a header defining the VkFormat enums as am I sure there will be many users of KTX2, e.g. WebGL/glTF, who will not want to include the full Vulkan header. Can such a file be generated from your JSON file?

You didn't answer my question about accepting this PR or waiting.

lexaknyazev:
The JSON file can be extended with VkFromat integer values if you think it's useful. I'd avoid merging now to reduce git diff noise wrt that JSON file.

MarkCallow:
Before you add integer values, We should see if such a header can be easily generated from the .xml file  Khronos uses as the base for generating the full Vulkan header.  That would make more sense.

lexaknyazev:
Looks like values are scattered across several XML sections because of versioning and extensions:
- **Core 1.0**
  - https://github.com/KhronosGroup/Vulkan-Docs/blob/master/xml/vk.xml#L4032
- **Core 1.1**
  - https://github.com/KhronosGroup/Vulkan-Docs/blob/master/xml/vk.xml#L7591
- **`VK_IMG_format_pvrtc`**
  - https://github.com/KhronosGroup/Vulkan-Docs/blob/master/xml/vk.xml#L8270
- **`VK_KHR_sampler_ycbcr_conversion`** (duplicated in Core 1.1)
  - https://github.com/KhronosGroup/Vulkan-Docs/blob/master/xml/vk.xml#L9528

My guess is that it's easier to parse an actual header:
https://github.com/KhronosGroup/Vulkan-Docs/blob/master/include/vulkan/vulkan_core.h#L568

MarkCallow:
Thanks for looking. They must have some code that puts the pieces together when they build vulkan_core.h but it might be hard to extract just that part of the scripts. You're probably right it is easier to parse vulkan_core.h. I think either of these is better than including the values in the JSON file as they will be easier to update or are you thinking of an automated way to add the values to the JSON?

lexaknyazev:
> are you thinking of an automated way to add the values to the JSON?

I don't think that this process could be fully automated - besides Vulkan enums, we'd need to add matching formats for other APIs.

MarkCallow:
I think it is better to have a separate header file declaring the VkFormat enums. Lets's generate it from vulkan_core.h as you suggested.

As for the JSON, I like to see it also include the blockSize or maybe blockSizeInBits and also whether the format is a depth or stencil format. The file could then be used to generate a switch with cases like

```
	        case VK_FORMAT_R4G4B4A4_UNORM_PACK16:
		case VK_FORMAT_B4G4R4A4_UNORM_PACK16:
		case VK_FORMAT_R5G6B5_UNORM_PACK16:
		case VK_FORMAT_B5G6R5_UNORM_PACK16:
		case VK_FORMAT_R5G5B5A1_UNORM_PACK16:
		case VK_FORMAT_B5G5R5A1_UNORM_PACK16:
		case VK_FORMAT_A1R5G5B5_UNORM_PACK16:
			pFormatSize->flags = VK_FORMAT_SIZE_PACKED_BIT;
			pFormatSize->paletteSizeInBits = 0;
			pFormatSize->blockSizeInBits = 2 * 8;
			pFormatSize->blockWidth = 1;
			pFormatSize->blockHeight = 1;
			pFormatSize->blockDepth = 1;
                       break;
```
which is very helpful for software handling KTX files. The possible values for `flags` are:

```
typedef enum VkFormatSizeFlagBits {
	VK_FORMAT_SIZE_PACKED_BIT				= 0x00000001,
	VK_FORMAT_SIZE_COMPRESSED_BIT			= 0x00000002,
	VK_FORMAT_SIZE_PALETTIZED_BIT			= 0x00000004,
	VK_FORMAT_SIZE_DEPTH_BIT				= 0x00000008,
	VK_FORMAT_SIZE_STENCIL_BIT				= 0x00000010,
} VkFormatSizeFlagBits;
```

We don't need `VK_FORMAT_SIZE_PALETTIZED_BIT`.

The above code comes from the manually generated https://github.com/KhronosGroup/KTX-Software/blob/master/lib/vk_format.h.

MarkCallow:
@lexaknyazev I'd like to get this PR completed and added to the spec as soon as possible. When do you think you can do the updates?

lexaknyazev:
Could we lay out all needed updates? Here's my understanding based on the comments above:
- Do not include integer enum values.
- Add some sort of channel info. Limit to depth/stencil, or expose all channels? In what form? Couple options (just for example):
  - ```json
    {
        "type": "RAW",
        "vkFormat": "VK_FORMAT_R8_UNORM",
        "typeSize": 1,
        "blockWidth": 1,
        "blockHeight": 1,
        "blockDepth": 1,
        "hasRed": true,
        "hasGreen": false,
        "hasBlue": false,
        "hasAlpha": false,
        "hasDepth": false,
        "hasStencil": false,
        "glInternalFormat": "GL_R8",
        "glDesktopOnly": false,
        "glFormat": "GL_RED",
        "glType": "GL_UNSIGNED_BYTE",
        "dxgiFormat": "DXGI_FORMAT_R8_UNORM",
        "mtlFormat": "MTLPixelFormatR8Unorm"
    },
  - ```json
    {
        "type": "PACKED",
        "vkFormat": "VK_FORMAT_D16_UNORM_S8_UINT",
        "typeSize": 4,
        "blockWidth": 1,
        "blockHeight": 1,
        "blockDepth": 1,
        "channels": "ds",
        "glInternalFormat": null,
        "glDesktopOnly": null,
        "glFormat": null,
        "glType": null,
        "dxgiFormat": null,
        "mtlFormat": null
    },
- Add block size in bytes.

---

We could also make a JSON form of DFD and include all its fields to this list instead of cherry-picking.

lexaknyazev:
Also, we've got a minor inconsistency in the current draft:
- Section "Depth and Stencil Formats"
  > `VK_FORMAT_D16_UNORM_S8_UINT` is defined as two 16-bit words per texel.
- Section "typeSize"
  >  For the combined depth/stencil formats using the layouts defined in this specification, the value will be 4.

Since I'm going to include `typeSize` for all known formats into JSON, we need to resolve this.

MarkCallow:
The only one of those flag bits I mentioned that I actually use in libktx is COMPRESSED_BIT. It's only used in the reader and a reader can get this information from the DFD, assuming we resolve issue #64 to still have a DFD. The channel information is also available from the DFD. The only items I think we need to add to the JSON are the block size and type size. My code only uses the block size in bytes. Either InBits or InBytes is fine. Also to avoid confusion with blockSize, blockWidth etc, should be called blockWidthInTexels or something like that.

I see 2 uses for the JSON: creating the mapping table and creating a switch statement for readers like the example I have above. Writers will need something to figure out the VkFormat to use given the format of whatever input files they are handling. That is something different from the JSON we have here.

MarkCallow:
I no longer think we need any additions to the JSON, Not flags, VkFormat integer values or typeSize. I'm generating a file with the format enums for my needs from vulkan_core.h. typeSize is in the header and can be calculated from the DFD which also has all the information conveyed by the flags.

Therefore please rebase this PR and add a commit to make the list collapsible. I'd like to get this in the spec as soon as possible.

lexaknyazev:
This should be good to go.

PR #60, opened 2019-02-20
Fix #14. Remove glTypeSize & endianness.

### LINKED ISSUE #14: How to handle type-awareness disparity between DFD and `VkFormat`

MarkCallow:


lexaknyazev:
`endianness` and `glTypeSize` fields were removed from the file header but their corresponding sections are still in the spec, is it intended?

Also the updated "Endianness" section in "General comments" looks strange with "XXX" and question marks.

MarkCallow:
Silly me. Sorry. It's missing a commit. I'll fix it in a moment.

MarkCallow:
It's fixed now.

PR #63, opened 2019-02-22
Describe depth-stencil formats

### LINKED ISSUE #24: Depth/stencil textures support
We need to verify whether KTX2 needs to support depth and/or stencil textures.

If it does:
- Which texture types could be used with D/S data?
- How does swizzling interact with D/S?

lexaknyazev:
Fixes #24.

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -400,6 +400,32 @@ specification as the formats are not in wide use and applications
 needing them can use KTX version 1.
 ====
 
+==== Depth and Stencil Formats
+Despite that Vulkan requires separate uploads of depth and stencil
+components, combined depth/stencil pixel formats can be used with KTX.
+[NOTE]
+.Rationale
+====
+Other GPU APIs support combined uploads and given KTX data alignment it's 
+trivial to upload components separately in Vulkan.
+====
+
+Depth or stencil formats cannot be used for 3D textures.
+
+`VK_FORMAT_D16_UNORM_S8_UINT` is defined as two 16-bit words per texel.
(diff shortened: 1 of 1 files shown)


ktxspec.adoc
```diff
@@ -400,6 +400,32 @@ specification as the formats are not in wide use and applications
...
+value in the eight LSBs and zeros in the eight MSBs.
+
+`VK_FORMAT_D24_UNORM_S8_UINT` is defined as one 32-bit word per texel
+with the S8 value in the eight LSBs of the word and the D24 value in the MSBs.
+
+`VK_FORMAT_D32_SFLOAT_S8_UINT` is defined as two 32-bit words per texel.
+The first word contains the floating-point D32 value. The second word contains
+the S8 value in the eight LSBs and zeros in the MSBs.
+
+`VK_FORMAT_S8_UINT`, `VK_FORMAT_D16_UNORM`, `VK_FORMAT_X8_D24_UNORM_PACK32`,
```
MarkCallow: Shouldn't we say the X8 bits should be 0?

lexaknyazev:
From https://github.com/KhronosGroup/KTX-Specification/pull/61#issuecomment-465592355:
> I think for all of them `typeSize` should be 4.

That would affect packing of `VK_FORMAT_D16_UNORM_S8_UINT`. We should decide whether it's one 32-bit word or two 16-bit words. In the latter case, `typeSize` would be 2.

ktxspec.adoc
lexaknyazev: Fixed.

MarkCallow:
Thank you.

Issue #64, opened 2019-02-24
DFD usage and implementation


lexaknyazev:
_The following feedback is based on DFD revision 1.2._

The DFD spec gives a lot of flexibility for providing formal definitions of "unknown" pixel formats such as uneven or padded bit widths, Bayer patterns, etc. Many DFD options duplicate fixed `vkFormat` definitions, so this leads to lots of implementation corner cases. 

DFD fields are:
#### color_model
```
    KHR_DF_MODEL_UNSPECIFIED
    KHR_DF_MODEL_RGBSDA
    KHR_DF_MODEL_YUVSDA
    KHR_DF_MODEL_YIQSDA
    KHR_DF_MODEL_LABSDA
    KHR_DF_MODEL_CMYKA
    KHR_DF_MODEL_XYZW
    KHR_DF_MODEL_HSVA_ANG
    KHR_DF_MODEL_HSLA_ANG
    KHR_DF_MODEL_HSVA_HEX
    KHR_DF_MODEL_HSLA_HEX
    KHR_DF_MODEL_YCGCOA
    KHR_DF_MODEL_YCCBCCRC
    KHR_DF_MODEL_ICTCP
    KHR_DF_MODEL_CIEXYZ
    KHR_DF_MODEL_CIEXYY
    KHR_DF_MODEL_BC1A
    KHR_DF_MODEL_BC2
    KHR_DF_MODEL_BC3
    KHR_DF_MODEL_BC4
    KHR_DF_MODEL_BC5
    KHR_DF_MODEL_BC6H
    KHR_DF_MODEL_BC7
    KHR_DF_MODEL_ETC1
    KHR_DF_MODEL_ETC2
    KHR_DF_MODEL_ASTC
```
Obviously, last ten options have 1:1 mappings with the corresponding compressed formats. Uncompressed formats can somewhat freely choose `color_model` from 16 other options and the application behavior is out of scope of KTX conformance. I think it's safe to assume that overwhelming majority of writers would use `RGBSDA` while most readers would ignore this filed altogether.

#### color_primaries
```
    KHR_DF_PRIMARIES_UNSPECIFIED
    KHR_DF_PRIMARIES_BT709 = KHR_DF_PRIMARIES_SRGB
    KHR_DF_PRIMARIES_BT601_EBU
    KHR_DF_PRIMARIES_BT601_SMPTE
    KHR_DF_PRIMARIES_BT2020
    KHR_DF_PRIMARIES_CIEXYZ
[...]

PR #75, opened 2019-04-25
Add limitations to DFD use from issue #64.

MarkCallow:


ktxspec.adoc
```diff
@@ -740,14 +740,28 @@ These 3 items combined form a _Data Format Descriptor_
...
 sufficient information for a KTX2 file to be parsed without having to
 refer to <<KDF13>>. It consists of one or more _Descriptor Blocks_
 (dfDescriptorBlock).
 
+The following restrictions must be obeyed when setting the fields of the
+DFD and the image data must match the chosen values:
+
+* `color_model` must be `KHR_DF_MODEL_RGBSDA`, `KHR_DF_MODEL_YUVSDA`
+  or one of the block compressed color models, currently
+  `KDF_DF_MODEL_BC1` to `KDF_DF_MODEL_ASTC`.
```
lexaknyazev: Are you going to update this statement after KDFS 1.3 release? I think we can put updated enums now.

ktxspec.adoc
```diff
@@ -740,14 +740,28 @@ These 3 items combined form a _Data Format Descriptor_
...
+* `color_primaries` may be any of the available values since conversion
+  to the display's primaries is relatively straightforward.
+  `KHR_DF_PRIMARIES_BT709/SRGB` is recommended for LDR images.
+  `KHR_DF_PRIMARIES_BT2020` is recommended for HDR images.
```
lexaknyazev: I don't think that LDR and HDR are the correct terms here. Also, listing "recommendation" between the restrictions may be confusing.

ktxspec.adoc
```diff
@@ -740,14 +740,28 @@ These 3 items combined form a _Data Format Descriptor_
...
+* `transfer_function` must be `KHR_DF_TRANSFER_LINEAR` or
+  `KHR_DF_TRANSFER_SRGB` since these are the only ones supported by the
+  vast majority of graphics hardware and correct filtering and blending
+  without hardware support is difficult to impossible.
+* If `transfer_function` is `KHR_DF_TRANSFER_SRGB` then
+  `<<_vkFormat,vkFormat>>` must be one of the `+*_SRGB{,_*}+` formats.
```
lexaknyazev: I think that this restriction must be bidirectional, i.e., sRGB `vkFormat` must always be used with `KHR_DF_TRANSFER_SRGB`.

ktxspec.adoc
MarkCallow: I'll do it now.

ktxspec.adoc
MarkCallow: What terms do you suggest? I don't think having recommendations here is confusing but do you have a suggestion as to where else to put this recommendation?

ktxspec.adoc
MarkCallow: It's implied but I'll make it explicit.

ktxspec.adoc
lexaknyazev: I'd omit this recommendation for now. Do we have a real-world use case for using BT.2020 primaries with GPU textures? 

I'd change "HDR" to "wide color gamut", since "HDR" could mean different things to different developers depending on the context.

ktxspec.adoc
```diff
@@ -740,14 +740,30 @@ These 3 items combined form a _Data Format Descriptor_
...
+  to the display's primaries is relatively straightforward.
+  `KHR_DF_PRIMARIES_BT709/SRGB` is recommended for LDR images.
+  `KHR_DF_PRIMARIES_BT2020` is recommended for HDR images.
+* `transfer_function` must be `KHR_DF_TRANSFER_LINEAR` or
+  `KHR_DF_TRANSFER_SRGB` since these are the only ones supported by the
+  vast majority of graphics hardware and correct filtering and blending
+  without hardware support is difficult to impossible.
+* If `<<_vkFormat,vkFormat>>` is one of the `+*_SRGB{,_*}+` formats then
+  `transfer_function` must be `KHR_DF_TRANSFER_SRGB`. For all other
+  formats it must be `KHR_DF_TRANSFER_LINEAR`
```
lexaknyazev: > For all other formats it must be `KHR_DF_TRANSFER_LINEAR`

Is compatible with Vulkan 1.1 multiplane video formats? Are we going to support them?

ktxspec.adoc
MarkCallow: > Do we have a real-world use case for using BT.2020 primaries with GPU textures?

Yes. As I wrote earlier, Gran Turismo Sport HDR version.

I think wide color gamut and high dynamic range are different potentially overlapping things. So I'm not sure "wide color gamut" is sufficient especially as I'm not sure you can use sRGB/BT.709 primaries for high dynamic range images.

ktxspec.adoc
MarkCallow: Good question. We'll have to find out the answer. I do want to support at least YUV formats.

ktxspec.adoc
lexaknyazev: After some more digging, I'm sure that this recommendation should be removed as workflows may be very different.
For example, here're some excerpts from NVIDIA's guidelines for working with HDR/WCG/UHD content:

> a HDR image with sRGB chromaticities can appear more richly colored than an SDR image with DCI-P3 primaries

> Given the complexity and cost of developing content for a game, it is generally impractical to expect UHD and sRGB assets, or convert over the art flow to generate most art in these new wider color spaces. 

> Here are the core principles:
>- Author content with sRGB primaries as today
>- Render high quality HDR data using physically-based shading
>- Perform all rendering operations with respect to the sRGB primaries you use today

lexaknyazev: More on this from NVIDIA:
> Rendering using the sRGB primaries is directly related to this. sRGB primaries are what you render with today. If you were to take your sRGB content you create today and convert it into colors described by the BT. 2020 primaries before performing lighting, you’d see much different results. This happens because material interactions are relative to the color space in which they are defined. The modulation operator we all use to compute the interaction between a material and a light source will produce different results if you change the primaries. As a game developer, you don’t want your artists authoring content to see an object looking brown on their monitor, but the same object being green or purple when you run the game on a UHD class display.

So, the ideal workflow would be to use BT.2020 primaries for all color interactions and display the result on a BT.2020-compliant display (AFAIK, the best consumer displays support only DCI-P3 for now). Obviously, actual workflows will be sub-optimal for quite a long time, so such recommendation ("use HDR with BT.2020") doesn't seem very useful.

ktxspec.adoc
MarkCallow: Done in commit https://github.com/KhronosGroup/KTX-Specification/pull/75/commits/6d5765e3fc392345d73f17f8ed9457d27ccc14b2.

ktxspec.adoc
MarkCallow: Removed primary recommendations in commit https://github.com/KhronosGroup/KTX-Specification/pull/75/commits/6d5765e3fc392345d73f17f8ed9457d27ccc14b2.

ktxspec.adoc
MarkCallow: For now I've excepted YUV formats and added a note that YUV support is not yet decided and the YUV section will be updated after a decision.

MarkCallow:
@lexaknyazev, please review the latest version.

ktxspec.adoc
```diff
@@ -744,14 +744,48 @@ These 3 items combined form a _Data Format Descriptor_
...
+The following restrictions must be obeyed when setting the fields of the
+DFD. The value of `<<_vkformat,vkFormat>>` and the image data must never
+contradict the DFD.
+
+* `color_model` must be `KHR_DF_MODEL_RGBSDA`, `KHR_DF_MODEL_YUVSDA`
+  or one of the block compressed color models listed in <<KDF13>> Section
+  5.6 or its successors, currently `KDF_DF_MODEL_BC1A` to
+  `KDF_DF_MODEL_PVRTC2`.
+* `color_primaries` may be any of the available values since conversion
+  to a display's primaries is relatively straightforward.
```
lexaknyazev: It is "relatively straightforward" only when the white point is the same. Otherwise it's a bit more complicated (from that NVIDIA guide):
> In addition to simply changing the color primaries, it may be necessary to adjust the white point of the image as well. The standard way of doing this is through a Chromatic Adaptation Transform (CAT). The common CAT solutions are the Bradford or Von Kries transforms. These convert the colors to an LMS color space, then scale by destination white / source white (also in LMS space), and finally convert back from LMS space. As the transform to LMS space in both Bradford and Von Kries is simply a 3x3 matrix multiply, the entire CAT can be collapsed into a single 3x3 multiply, just like the transform of primaries.

Do matrices provided in the DFD spec accommodate for this?

ktxspec.adoc
```diff
@@ -744,14 +744,48 @@ These 3 items combined form a _Data Format Descriptor_
...
+
+[NOTE]
+====
+Specifications which refer to KTX2 as their base format may wish to
+restrict the `color_primaries` that can be used.
+
+`KHR_DF_PRIMARIES_BT709/SRGB` is recommended for low dynamic range,
+standard gamut images.
```
lexaknyazev: I think that IBL use case may be an exception here, since it uses values outside of `0 .. 1` range with sRGB primaries.

ktxspec.adoc
MarkCallow: It boils down to a single matrix multiply, which of course the hardware is very good at doing, that converts both primaries and white point. This single matrix multiply is what I mean by relatively straightforward. Can you propose some alternative language. I certainly don't want to get into details of color space conversion here.

I'll check the matrices in the DFD spec. In any case the color spaces are all standard which proper specifications and ICC profiles which will provide the necessary information.

ktxspec.adoc
MarkCallow: I will add this exception when I add info about RGB[DEM] formats. I'll make a note in issue #80 to remind me.

ktxspec.adoc
MarkCallow: I think the matrices in KDFS spec. are for von Kries transform but, given the nature of the KDFS spec., I think the chapter 14 introduction is explanatory. I do not think that KTX2 needs to specify what CAT should be used. If glTF starts to allow different primaries it may want to mandate a CAT. HTML5 should probably specify the CAT to be used by browsers when converting from sRGB/BT.709 primaries to the display primaries.

I'll change the language to something along the lines of 

> ... conversion of the primaries and white point to those of a display's can be done by a 3x3 matrix multiply.

lexaknyazev: > HTML5 should probably specify the CAT...

If only web standards were so specific...

Rendering of HTML Images with embedded ICC profiles is up to implementation, Canvas API should someday consistently support rendering to float16 surfaces (IIUC, with extended sRGB).

I'm fine with the latest proposed language.

PR #83, opened 2019-05-16
Require cubemap faces to be squares

lexaknyazev:


MarkCallow:
Thanks for spotting this. I thought we already had this restriction documented. libktx certainly enforces it.

Rather than me making a new PR, please update the PR to fix a longstanding typo in the numberOfFaces section (3.6 if you're looking at the output doc.). The third sentence begins "Cube map faces". Please change it to "Cubemap faces" because we use "cubemap" everywhere else in the document.

lexaknyazev:
Done.

Issue #85, opened 2019-05-21
Change offset and length names to match glTF


MarkCallow:
glTF will mirror part of the KTX2 header in a JSON schema. glTF has a convention to use byteOffset and byteLength as the names of the related items. For consistency and avoidance of confusion it would be good to use the same names within the KTX2 spec. Since these names do not appear in the KTX2 file, changing is not a problem.

So change offset to byteOffset and bytesOfXXX to XXXByteLength.

Also refer to the glTF KTX2 as an example of providing a separate KTX2 header.

lexaknyazev:
For reference, [current JSON-Schema draft](https://github.com/KhronosGroup/glTF/blob/khr_ktx2_ibl/extensions/2.0/Khronos/KHR_image_ktx2/schema/image.KHR_image_ktx2.schema.json).

MarkCallow:
How can you get GitHub to close issues automatically when the PR is merged?

lexaknyazev:
@MarkCallow 
https://help.github.com/en/articles/closing-issues-using-keywords

MarkCallow:
arrayElementCount seems awkward alongside faceCount and levelCount. I'd like to change it to layerCount. What do you think @lexaknyazev?

lexaknyazev:
Sounds good. Vulkan also calls them layers.

PR #87, opened 2019-05-29
Rename header fields to match glTF conventions and KTX2 extension schema

MarkCallow:
For issue #85. Please review, especially @lexaknyazev.

One problem is that `numberOfMipLevels` has been renamed `levels` and the level index array is also called `levels`.  I'd prefer to keep the level index array as `levels`. Can we rename levels to, e.g., `mipLevels` in both KTX2 and the glTF schema?

Another issue is that, because some of the names are so generic, it will be much harder to rename these fields again. For example "faces" is used in many places where is does not mean the field containing the _number_ of faces.

And another is software like libktx that will support both versions of KTX. I'll probably make some macros so either the KTX1 or KTX2 name can be used for fields that occur in both versions.

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -127,9 +127,9 @@ UInt32 typeSize
 UInt32 pixelWidth
 UInt32 pixelHeight
 UInt32 pixelDepth
-UInt32 arrayElements
-UInt32 faces
-UInt32 levels
+UInt32 numberOfArrayElements
+UInt32 numberOfFaces
+UInt32 numberOfMipLevels
 UInt32 supercompressionScheme
 UInt64 byteLength
 UInt64 byteLengthUncompressed
@@ -146,7 +146,7 @@ struct {
     UInt64 byteOffset
     UInt64 byteLength
     UInt64 byteLengthUncompressed
-} levels[max(1, levels)]
+} levels[max(1, numberOfMipLevels)]
 
 // Data Format Descriptor <3>
 UInt32 dfdTotalSize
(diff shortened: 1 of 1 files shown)


lexaknyazev:
> Can we rename levels to, e.g., `mipLevels` in both KTX2 and the glTF schema?

Sure.

My main concern was with `byteLength` and `byteOffset`. All other names can be freely adjusted.

Another issue is that since JSON natively supports arrays, there're usually no properties that denote "array length". So we don't really have a convention how such fields should be named: "numberOfThings", "numThings", "thingsCount", "things", etc. A possible way to consistently resolve this is to always put the most descriptive noun last, i.e. change to something like `arrayElementsCount`, `facesCount`, and `levelsCount`.

MarkCallow:
I'm tempted to remove `imagesByteLength` and `imagesByteLengthUncompressed`. I put them there so an app could allocate the entire memory up front. However in almost all cases app's will need the level index so having to iterate over the index adding up the sizes is not much more complex. Opinions?

I will revert the changes to fields denoting array length. I'll probably use something along the lines @lexaknyazev suggested to be consistent with the `byte{Offset,Length}` nomenclature.

lexaknyazev:
> I'm tempted to remove `imagesByteLength` and `imagesByteLengthUncompressed`. 

👍

MarkCallow:
I've made all the changes except removal of `imagesByteLength{,Uncompressed}`. I'll await further feedback on the removal.

lexaknyazev:
Nit: maybe `byteLengthUncompressed` to `uncompressedByteLength`.

MarkCallow:
Fixed #85.

PR #91, opened 2019-08-01
Fix broken links.

sevmeyer:
Hello

I'm quite fond of KTX, and I'm glad the standardization effort continues with KTX2. In the current draft, I noticed some of the Vulkan links are broken because the anchor names have changed.

This commit fixes all broken links I found.


sevmeyer:
The commit seems to break two cross references, I apologize. I think the cause is my outdated Asciidoctor version, which flagged the references as invalid.
I will fix the commit and reopen a new merge request.

PR #92, opened 2019-08-01
Fix broken links.

sevmeyer:
This time it should be correct.

In the current draft, I noticed some of the Vulkan links are broken because the anchor names have changed. This commit fixes those links, as well as two OpenGL links that were missing a slash.


MarkCallow:
Thank you for the fixes and your kind words about KTX.

PR #97, opened 2019-08-06
Add ASTC HDR & 3D enums

### LINKED ISSUE #84: Update spec. when ASTC HDR & 3D Vulkan enums are available
Remove comments about using the DFD to distinguish these formats.

lexaknyazev:
This adds HDR and 3D enums to the formats mapping list. Fixes #84.


lexaknyazev:
@MarkCallow PTAL.

MarkCallow:
Thanks for driving this issue and fixing it.

formats.json
```diff
@@ -2141,6 +2323,440 @@
...
+        "glInternalFormat": "GL_COMPRESSED_RGBA_ASTC_12x12_KHR",
+        "glDesktopOnly": false,
+        "glFormat": null,
+        "glType": null,
+        "dxgiFormat": null,
+        "mtlFormat": null
+    },
+    {
+        "type": "ASTC",
+        "vkFormat": "VK_FORMAT_ASTC_3x3x3_UNORM_BLOCK_EXT",
```
mosra: @lexaknyazev apologies if this is not the right place to ask -- but where are those 3D Vulkan enums defined (or proposed, at least)? I'm investigating 3D ASTC support in various APIs and the only places where I found the definitions are the `OES_texture_compression_astc` extension for GL and this PR for Vulkan. The current Vulkan spec doesn't have those enums and the only related issue I found is https://github.com/KhronosGroup/Vulkan-Docs/issues/926, but there isn't any mention about them either.

MarkCallow: They come from a draft Vulkan extension within Khronos, which will be released once there are 2 implementations. When that might happen, I don't know. The enum values will not change.

mosra: Wonderful, thanks for clarifying that :)

PR #103, opened 2019-11-02
Supercompression update

### LINKED ISSUE #70: Support GIF-style animation?
Should we make changes to allow a .ktx2 files to contain an animation, i.e indicate that the set of images is to be displayed sequentially?

An immediate question I have is why do people still use animated GIFs vs. some video file format? Is it about lack of video authoring tools or difficulty dealing with cross-platform video on the Web or something else?

This would be a ktx2 file with  levels=1, faces=1 and layers=<number of frames>. Support is probably just a case of adding suitable metadata that indicates the layers should be displayed sequentially and specifying the interval between frames.

If the sequences are expected to be short, one can even imagine uploading the entire contents to the GPU as an array texture, redrawing at the appropriate interval, incrementing the third texture coordinate in between. It could probably be done entirely on the GPU.

### LINKED ISSUE #86: How to carry the Basis bitstream?
We have to figure out how to indicate the texture data is using Basis supercompression and can be transcoded.  We'll likely have to use VK_FORMAT_UNKNOWN, because inflation and transcoding are done together. If they were done separately post-inflation would be ETC1S. IIUC as it is now you have to tell the transcoder you want ETC1 in order to get the ETC1S output.

The DFD will probably just provide the color model, primaries and transfer function and have no sample info.

So we need something else to identify it as Basis. Is identifying the supercompressionScheme as Basis enough?

### LINKED ISSUE #89: Make supercompression schemes optional, except Basis
Following WG discussion the consensus is to make all except Basis optional. Subsidiary question is what optional schemes to include: LZMA only, LZMA & Zstd, LZMA, Zstd  and Deflate?

LZMA has a higher compression ratio than Zstd but its decompression is slow (and reportedly much slower than Zstd).

### LINKED ISSUE #98: Basis information in the spec is outdated
Specifically, the spec repeatedly refers to _Crunch CRN_ rather than _Basis_ and should be updated to reflect the more recognizable public name.  Additionally, there is a note in 3.8.1 that reads 

>  A file that specifies Crunch CRN with base formats other than ETC, ETC2 and BC[1-3] (S3TC_DXT[1-5]) must be considered invalid.

It looks like base format is a reference to a no longer extant field, unless I'm missing something.

### LINKED ISSUE #100: Where should we indicate the number of channels in BasisU data?
Basis Universal compressed data may have 2, 3 or 4 channels. The number of channels affects the choice of transcode target format so the recipient of a file needs to know the information. This could be provided within the BasisU global data or by the DFD. Which should we use?

Currently presence of an alpha channel (alpha slices) is indicated by a flag in the global data similar to the .basis format. There is nothing to indicate if there are only 2 channels either in global data or in .basis files.

From an API user's point of view, having a function on the ktxTexture object that returns the number of channels, regardless of the internal format of the ktxTexture data is clean and simple. From an implementer's point of view having that function interact with only one part of the file is simpler. So I lean toward using the DFD.

MarkCallow:
This introduces a change incompatible with previously generated Basis compressed .ktx2 files in order to store the number of components of the original image after Basis supercompression. The information is retained in the DFD.

Fixes #70. Fixes #86. Fixes #89. Fixes #98. Fixes #100

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -533,14 +533,15 @@ of `0` indicates no supercompression.
 [[supercompressionSchemes]]
 .Supercompression Schemes
 |===
-| Scheme Id            | Scheme Name     | Level Data Format | Global Data Format
-| 0                    | None            |   n/a             | n/a
-| 1                    | Basis Universal |   T.B.C           | <<basisu_gd,BasisU Global Data>>
-| 2                    | ZLIB            |  <<ZLIB>>         | T.B.D
-| 3                    | Zstandard       |  <<ZSTD>>         | T.B.D
-| 4･･･0xffff           | Reserved^1^     |                   |
-| 0x10000･･･0x1ffff    | Reserved^2^     |                   |
-| 0x20000･･･0xffffffff | Reserved^3^     |                   |
+| Scheme Id            | Scheme Name     | Level Data Format  | Global Data Format
+| 0                    | None            |   n/a              | n/a
+| 1                    | Basis Universal |   T.B.C            | <<basisu_gd,BasisU Global Data>>
+| 2                    | ZLIB            |  <<RFC1950>>       | T.B.D
+| 3                    | Zstandard       |  <<RFC8478>>       | T.B.D
+| 4                    | LZMA            |  <<LZMA>>          | T.B.D
+| 5･･･0xffff           | Reserved^1^     |                    |
+| 0x10000･･･0x1ffff    | Reserved^2^     |                    |
+| 0x20000･･･0xffffffff | Reserved^3^     |                    |
 |===
 
 [small]
@@ -603,7 +604,7 @@ not be visible.
 
 ===== ZLIB
 
-* With Deflate <<DEFLATE>> compression scheme.
+* With Deflate <<RFC1951>> compression scheme.
 
 ===== Zstandard
 * Only _Zstandard_ frames are required. Inflators may skip _Skippable_
@@ -1538,18 +1539,14 @@ Therefore there is no advantage to storing it there.
   {url-khr-reg}/OpenGL/extensions/OES/OES_texture_compression_astc.txt[GL_OES_texture_compression_astc].
 Sean Ellis, et al. The Khronos Group, July 2016.
 
-////
-// "L." after the doc. title is to make the correct author name
-// L. Peter Deutsch. If I put it at the start of the line following
-// the title, Asciidoctor thinks I am trying to make a list.
-////
-- [[[DEFLATE]]] https://tools.ietf.org/html/rfc1951[DEFLATE Compressed
-  Data Format Specification version 1.3 (RFC1951)]. L.
-Peter Deutsch. IETF Network Working Group, May 1996.
-
 - [[[KDF13]]] {url-khr-reg}/DataFormat/specs/1.3/dataformat.1.3.html[Khronos
   Data Format Specification 1.3].
 Andrew Garrard. The Khronos Group.
+
+- [[[LZMA]]] https://www.7-zip.org/a/lzma-specification.7z[LZMA
+  Specification (DRAFT version)] and
+  https://www.7-zip.org/sdk.html[LZMA Software Development
+  Kit]. Igor Pavlov. June 2015.
 
 - [[[OESCPT]]] {url-khr-reg}/OpenGL/extensions/OES/OES_compressed_paletted_texture.txt[GL_OES_compressed_paletted_texture].
 Aaftab Munshi. The Khronos Group, July 2003.
@@ -1574,10 +1571,27 @@ Imagination Technologies Limited, 2011
  ECMA-262 5.1{nbsp}Edition, Section 15.10: RegExp (Regular Expression) Objects].
 Ecma International, June 2011.
 
-// "S" after doc title avoids the Asciidoctor list issue.
+////
+// "L." after the doc. title is to make the correct author name
+// L. Peter Deutsch. If I put it at the start of the line following
+// the title, Asciidoctor thinks I am trying to make a list.
+////
+- [[[RFC1950]]] https://tools.ietf.org/html/rfc1950[ZLib Compressed Data
+  Data Format Specification version 3.3]. L.
+Peter Deutsch, Jean-Loup Gailly. IETF Network Working Group, May 1996.
+
+// "L.", "S." and "Y." after doc titles avoid the Asciidoctor list issue.
+- [[[RFC1951]]] https://tools.ietf.org/html/rfc1951[DEFLATE Compressed
+  Data Format Specification version 1.3]. L.
+Peter Deutsch. IETF Network Working Group, May 1996.
+
 - [[[RFC2119]]] https://www.ietf.org/rfc/rfc2119.txt[Key words for use in RFCs to
 Indicate Requirement Levels]. S.
 Bradner. IETF Network Working Group, March 1997.
+
+- [[[RFC8478]]] https://tools.ietf.org/html/rfc8478[Zstandard Compression
+and the application/zstd Media Type.]. Y.
+Collet, M. Kucherawy, Ed. Internet Engineering Task Force (IETF), October 2018.
 
 - [[[VULKAN11]]] {url-khr-vulkan}/specs/1.1/html/vkspec.html[Vulkan^®^
 1.1 - A Specification].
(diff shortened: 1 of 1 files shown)


MarkCallow:
I will take the absence of comments as approval of these changes. I will wait until Monday morning, JST, before merging to give everyone a final chance.

ktxspec.adoc
```diff
@@ -837,7 +897,178 @@ individual `keyAndValueByteLength` fields.
...
+codebook in the form of supercompressed ETC1S endpoint and selector
+data and the huffman tables that are core to lossless supercompression.
+
+It also contains an array of _image descriptors_ with flags for each image
+and the offset and length within the image's mip level of the data for
+that image. Images may consist of 1 or 2 _slices_ depending on the
+components of the pre-deflation image as detailed below.
+
+_One-component images_ must have the component value replicated in all 3
+components of the _rgb_ slice of the compressed data and have no _alpha_
```
lexaknyazev: Given that slice data is generated by Basis Universal encoder from an unknown source data, it looks like the KTX2 spec technically disallows ambiguous source/target combinations. For example:
- One-component image can be encoded to BasisU with RGB channels populated with the same value. Transcoding it to BC1/ETC1: three RGB slice channels populating three RGB texture sampler channels with the same value.
- One-component image can be encoded to BasisU with only green channel populated with zeros in red and blue . Transcoding it to BC1/ETC1: three RGB slice channels populating three RGB texture sampler channels with data only in green.

This (and subsequent) restrictions are added to potentially reduce endpoints footprint and to ensure consistency of transcoded images when the runtime transcoded format is uncompressed or suboptimal (otherwise 1- and 2-component images could always use green channel), right?

ktxspec.adoc
```diff
@@ -837,7 +897,178 @@ individual `keyAndValueByteLength` fields.
...
+
+Basis Universal supports inter-frame (video) encoding for 2D slices.
+If `isIFrame` is set the image (frame) is an I frame. That is, it
+does not refer to the previous image. If unset, the image is a P
+frame. If the file is not a _video_ file it must be unset.
+
+_Video_ files have a <<_layercount,`layerCount`>> > 0 and must have
+KTXanimation metadata. See <<KTXanimation>>. The value of
+<<_layercount,`layerCount`>> is the number of frames in the video,
+i.e. layers become the temporal axis.
```
lexaknyazev: Since `KTXanimation` metadata could be used without supercompression, should the definition of _video_ file be put outside of BasisU section?

ktxspec.adoc
```diff
@@ -553,35 +558,90 @@ in the reference given in the _Level Data Format_ column of
...
+32+^| *_sampleLower:_* 0
+32+^| *_sampleUpper:_* 0
+|================================
+////
+
+[NOTE]
+.Rationale
+====
+The Basis Universal encoder combines encoding to ETC1S with deflation.
+The transcoder combines inflation to ETC1S with transcoding to one
```
lexaknyazev: KDFS has a normative definition of ETC1S (as a strict subset of ETC1), while BasisU seems to diverge from it (at least wrt endpoints encoding). In my opinion, KTX2 should not call that custom BasisU encoding as just "ETC1S".

ktxspec.adoc
MarkCallow: Yes I want to restrict the possibilities so as to make behavior consistent and predictable and to reduce endpoint footprint. I'm basically following the undocumented behaviour of the BasisU encoder. It works only on 4 component images and uses lodepng to load png files. When converting to RGBA lodepng replicates "grey" to R, G & B for 1 and 2 component images. The BasisU encoder uses the RGB values unchanged. `libktx` does similar when preparing in memory RGBA images for the encoder. The encoder copies the alpha value to all 3 channels of the alpha slice.

Note that the transcoder does not currently have an option to transcode to uncompressed 1 or 2 component images only 3 or 4 component.

ktxspec.adoc
MarkCallow: I thought about this when writing it. I agree but could not decide on a good place to put it. How about in section 4 "General Comments"?

ktxspec.adoc
MarkCallow: The definition in KDFS is supposed to be what Rich calls ETC1S. We did run the definition by Rich when we added it to KDFS. Has something changed since? If so, we need to update KDFS. There is no other use for ETC1S beyond BasisU.

ktxspec.adoc
MarkCallow: It may be these particular restrictions, based on behaviour inherited by the BasisU encoder from lodepng, are not the right restrictions. In PNG, 1 & 2 component images are labelled grey and grey-alpha so swizzling the "grey" component to r, g & b makes sense. In KTX, OpenGL & Vulkan 1 & 2 component images are known as R & RG. I think the restrictions outlined make sense in this case too but I am open to alternatives.

Another issue is that currently the encoder supports a `separateRGToRGB_A` option. This causes the R & G components of an RGB{,A} input image to be swizzled to RGB and A respectively.  This option is copied from BasisU. I am not sure there is any utility to it. Is there any content creation pipeline that creates tangent space normal maps in 3 or 4 component textures? Those being the the advertised use of the option. [Note: if you were to pass a 2 component .png file to the BasisU encoder and set this option, the result would be wrong. The "grey" component would end up in all 3 components of the RGB *and* alpha slices.]

lexaknyazev: >  In PNG, 1 & 2 component images are labelled grey and grey-alpha so swizzling the "grey" component to r, g & b makes sense. In KTX, OpenGL & Vulkan 1 & 2 component images are known as R & RG.

Legacy versions of OpenGL had Luminance and Luminance-Alpha (1- and 2-component) formats that semantically match PNG. I think that the current approach is correct.

> This causes the R & G components of an RGB{,A} input image to be swizzled to RGB and A. This option is copied from BasisU. I am not sure there is any utility to it. 

Original 2-component image may be authored in JPEG which supports only 1 or 3-component images. Also, it could be an RG texture saved in PNG verbatim (with zeros in blue channel). This encoder option gives some additional flexibility, but maybe the right approach is to generalize input data swizzling so this option would translate to something like `--swizzle-inputs rrrg`.

ktxspec.adoc
lexaknyazev: Back then (before BasisU became open-source), we expected to use ETC1S RDO with some generic lossless step. So the KDFS definition of ETC1S is correct.

The issue is that BasisU is similar to but not exactly ETC1S.

ktxspec.adoc
lexaknyazev: Seems fine. Probably right after the table with texture types.

ktxspec.adoc
MarkCallow: Okay. Implementation of the option is already by a generalized swizzle. I think exposing that is a good idea. But that is a matter for `libktx` and `toktx` not the spec. Specwise I'll leave this as is but add some words so offering a swizzle option doesn't conflict.

ktxspec.adoc
MarkCallow: Will do.

ktxspec.adoc
MarkCallow: As I pointed out, BasisU is still refers to what it is using as ETC1S. We should update KDFS as the purpose of adding ETC1S was to describe what we would be doing in KTX2.

What exactly are the differences?

lexaknyazev: Currently, BasisU is a software product that uses "ETC1S" in its source code. Specwise, the KTX2 should refer to the BasisU bitstream (when it's available).

BasisU stores all endpoints and selectors in global tables; image slices contain only pointers to that data. Moreover, all data blocks are additionally losslessly compressed. To get GPU-ready ETC1S payload (even without transcoding), the decoder needs to decompress Huffman and to reconstruct the blocks. The last step is not documented anywhere yet.

KDFS ETC1S reference is useful for understanding the compression technique in general but certainly not enough to implement the codec.

MarkCallow: I think we are at cross purposes here. By referring to ETC1S I am not attempting, or claiming, to provide enough info to implement a codec. I am attempting to provide background as to why there are potentially 2 slices. I know steps are needed to reconstruct the ETC1S blocks and I know we need to refer to the BasisU bitstream, once available. There is already a place for the reference in the table of supercompression schemes.

It sounds like your earlier statement that "BasisU seems to diverge from it (at least wrt endpoints encoding)." is referring to BasisU putting the endpoints in the global table. Is that so or is there some other difference?

I will attempt rewording to make it clear the ETC1S is a building block and that further info is required to make a complete codec. I will include another T.B.D reference to the BasisU bitstream spec.

MarkCallow: Please take a look at 385a6b6.

lexaknyazev: The reworded text LGTM.

PR #104, opened 2019-11-11
Forbid empty and duplicate keys.

### LINKED ISSUE #94: Key/Value ambiguities (section 3.11)
There are some edge-case ambiguities for the key/value metadata (section 3.11), which might be worth to clarify.

#### Empty key

As I understand it, a key can be empty (`keyAndValue` begins with `0x00`). Is this a good idea? I don't expect empty keys to break any implementation, but they could be quite confusing.

#### Duplicate keys

Currently, the spec does not say what should happen if a key is included multiple times, possibly with different values. The pragmatic approach is to require that keys must be unique per file. Otherwise, there is a need for precedence and sorting rules.

MarkCallow:
Fixes #94.

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -856,6 +856,8 @@ or which is not modified by the user.
 Key/value data must be written to the file sorted by the Unicode
 code points of the keys starting from a key's first character.
 
+Keys must not appear more than once.
+
 ==== keyAndValueByteLength
 The number of bytes of combined key and value data in one key/value
 pair. This includes the size of the key, the required NUL byte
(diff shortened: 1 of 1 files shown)


ktxspec.adoc
```diff
@@ -866,6 +868,9 @@ that reads KTX files must not assume that value fields are NUL
 terminated). `keyAndValueByteLength` does not include the bytes in
 `<<_valuepadding,valuePadding>>`.
 
+`keyAndValueByteLength` must be at least 3, that is a 1 byte key plus
+its NUL terminator and the value's NUL terminator.
```
lexaknyazev: At one point, we had a metadata entry without value (just key). Given that the spec allows non-string values, NUL terminator is optional. So the minimum possible length is 2.

MarkCallow: Oops! Fix in PR 105.

PR #117, opened 2020-01-31
Change mip padding to support mmapping. Fixes #116.

### LINKED ISSUE #116: Change mip level padding requirements to support mmapping
There is a desire to make sure mip levels are aligned to meet the needs of mmapping. See https://github.com/KhronosGroup/KTX-Software/issues/161 for initial discussion.

The proposal is to require the first mip level in the file to be aligned on a texel block-sized boundary. For block-compressed textures subsequent mip levels will naturally fall on block-sized boundaries so `mipPadding` will be always be 0 bytes.

For uncompressed textures, it is necessary to be cognizant of the Vulkan requirement for bufferOffset in a VkBufferImageCopy to be both a multiple of 4 and of the texel block size. This means that 1, 2 and 3 component uncompressed textures may still require `mipPadding`.

If this is spec. change is acceptable, the trick now is to find a way to express all this simply in the specification. One complexity to be expressed is that padding following key/value data has to be `align(8)` if followed by supercompressionGlobalData, otherwise it has to be `align(texel block size)`.

MarkCallow:


ktxspec.adoc
```diff
@@ -194,20 +192,22 @@ After inflation from supercompression or when `supercompressionScheme ==
 .levelImages Structure
 [source, c]
 ----
+align( lcm(texel block size, 4) ) mipPadding <1>
```
zeux: This line implies that the alignment is 4 for supercompressed textures, but I think the desired behavior (spelled out below) is that there's no alignment. Should this say "sgdByteLength > 0 ? align(...)"?

ktxspec.adoc
```diff
@@ -1224,9 +1238,35 @@ level. The offset of a level's `levelImages` is provided by the
 When `<<_supercompressionscheme,supercompressionScheme>> != 0` these
 bytes are formatted as specified in the scheme documentation.
 
-=== mipPadding
-`mipPadding` is between 0 and 7 bytes of value `0x00` to make sure that all
-mip level data starts at a file offset that is a multiple of 8.
+==== mipPadding
+`mipPadding` is between 0 and stem:[lcm(texel\ block\ size, 4) -
```
zeux: What does "between" mean here? It reads as if it's optional but I'm assuming that it's required to be the specified number of 0 bytes.

zeux: ... Oh, sorry, I get it now. Ignore this.

ktxspec.adoc
MarkCallow: Line 195 is in the levelImages structure section about which the immediately preceding text says 

>After inflation from supercompression or when `supercompressionScheme ==
0`, `levelImages` looks like this:

PR #122, opened 2020-04-15
Fix typos and inconsistent type names.

warrenm:


MarkCallow:
Thank you for the fixes.

PR #148, opened 2021-04-13
Move cubemap coord system details to appendix.

MarkCallow:


ktxspec.adoc
```diff
@@ -2237,10 +2213,47 @@ particular version of OpenGL {,ES} or Vulkan.
...
+
+<<orientationFigure>> shows graphically how the cubemap images should be
+arranged.
+
+[[orientationFigure]]
+.Cubemap Coordinate System
+[.right.text-center]
+image::cubemap_coord_system.png[]
+
+If using a skybox to render the cubemap the uvw  coordinates passed to the
```
lexaknyazev, line 2238: For consistency, consider replacing `uvw` with `(s, t, r)` (as in OpenGL) or with `(s, t, p)` (as in GLSL).

ktxspec.adoc
```diff
@@ -2237,10 +2213,47 @@ particular version of OpenGL {,ES} or Vulkan.
...
+cubemap sampler need to match the KTX cubemap coordinate system, that is
+left-handed with +Y up.
+
+If using OpenGL's default clip-space, which is right-handed, you
```
lexaknyazev, line 2242: OpenGL's default clip-space is left-handed. It's the object space (before eye and projection transforms) that is usually considered right-handed.

lexaknyazev (commented):
We should also update the `KTXorientation` metadata description to say that it applies to all faces and replace "2D" with "2D or Cubemap" in the Type / Format table.

lexaknyazev:
To ensure ecosystem integrity wrt `KTXorientation`, we should provide a set of sample cubemaps that contain the same data pre-flipped with all 4 orientation metadata options. So that all four cubemaps look exactly the same when loaded and rendered correctly. The per-face content should include the name of the face (e.g. +Y) and its direction (e.g. UP).

MarkCallow:
> we should provide a set of sample cubemaps

That will have to wait until post release.

ktxspec.adoc
MarkCallow: Done.

ktxspec.adoc
MarkCallow: Fixed.

ktxspec.adoc
```diff
@@ -562,36 +563,11 @@ Refer to <<_texture_type>> for more details about valid values.
...
-the displayed scene being a mirror image of reality, a common error in
-samples found on the web.
-
-While Vulkan has a left-handed coordinate system, +Y is down, therefore
-skybox coordinates must be rotated 180 degrees around the X axis before
-they are used to sample a cubemap. Failure to do this will result in the
-scene being upside down.
-====
+Cubemap faces are stored in the order: +X, -X, +Y, -Y, +Z, -Z in a
+left-handed coordinate system with +Y up. All faces must have the
```
lexaknyazev, line 567: Slightly opinionated: it's not enough to define +Y as "up" to set the "ground truth" for a cubemap that is supposed to represent a real environment. One more axis needs to be tied (front/back/left/right).

MarkCallow: Please suggest modified language to address this. I don't have time for to and fro. I'm supposed to produce the release version of the spec. today and it is already 8:30PM.

ktxspec.adoc
```diff
@@ -2237,10 +2213,47 @@ particular version of OpenGL {,ES} or Vulkan.
...
+[appendix#cubemapOrientation]
+== Cubemap Orientation
+
+CAUTION: This appendix is non-normative.
+
+The KTX cubemap coordinate system in <<faceCount>> is directly
+compatible with the Vulkan and OpenGL cube samplers described by the
+face selection tables and equations for calculating (s, t) in
+{url-khr-vulkan}/specs/1.2-extensions/html/vkspec.html#_cube_map_face_selection_and_transformations[section
+15.5.3 Cube Map Face Selection and Transformations] of <<VULKAN12>>
```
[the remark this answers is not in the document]
MarkCallow: That reference is to the no-extensions version of Vulkan so the section number is unlikely to change again. I've updated it to the correct value. My comment earlier was talking about the situation with the version with all the extensions.

MarkCallow:
> We should also update the KTXorientation metadata description to say that it applies to all faces and replace "2D" with "2D or Cubemap" in the Type / Format table.

I have updated the metadata description. By Type / Format table are you talking about the _Texture Type_ table? I don't see why the "2D" in there should be replaced.

lexaknyazev:
> By Type / Format table are you talking about the _Texture Type_ table?

I refer to the table located in `KTXorientation` section.

ktxspec.adoc
lexaknyazev: On it. Will check with glTF's axises.

MarkCallow:
Thanks. I've updated that table. Everything is done except for defining one more axis. Waiting for language from you.

ktxspec.adoc
lexaknyazev: glTF defines its object space as: right-handed, **+Y up**, **+Z forward**, **+X left**.

Given that KTX cubemaps are left-handed, we have two options:
- +Z forward, +X right. In that case, transformation would be multiplying X by -1.
- -Z forward, +X left. In that case, transformation would be multiplying Z by -1.

MarkCallow: My cubemap samples have +Z forward and +X on the right. However I get the correct result in OpenGL by multiplying the cube coordinates' Z by -1. In Vulkan I get the correct result by rotating the cube's coordinate system around the X axis to because Vulkan has +Y down.

Wait a minute. What coordinates are you talking about? Are these cubemap face coordinates or skybox cube coordinates? I'm talking about the face coordinates.

MarkCallow: And if we're talking face coordinates there is no difference between your 2 cases.

lexaknyazev: In my opinion, the first option is better as it seems to be more popular across tools.

MarkCallow: +X left, seriously?

MarkCallow: My samples are using the GL RH-coord system. That must be why I have to multiply Z by -1.

What language do I need to modify? I think I just need to say +Z forward, +X on the right. The necessary transformation depends on the API or glTF or whatever. Is that correct?

MarkCallow: What is "forward" here? Does it mean in front of you so  going into the screen, i.e, -Z direction in OpenGL default CS? So glTF has +Z into the screen.

MarkCallow: @lexaknyazev please reply. I need to wrap this up.

lexaknyazev: > What coordinates are you talking about?

From the point of view of an observer that is located inside the cube, +Y is up, +X is on the right, and +Z is forward. That's the proposal. Alternative is to make -X right and -Z forward.

lexaknyazev: For the reference, here's glTF's object space:

lexaknyazev: > I think I just need to say +Z forward, +X on the right. 

Yeah, that's enough.

MarkCallow: Inside the cube. That's what I thought you were talking aboutl. The object space you picture is the same as OpenGL.

It seems to me the 2 alternatives are the same. If you are looking at the posz image and you have the posx image on your right, if you turn around so you are looking at the negz image the posx image will now be on your left.

So why is my sample working when I multiply z by -1? 

I have to give up tonight. I can't think properly any more.

As I asked earlier please propose modifications to the existing language.

MarkCallow:
I figured out the disconnect. Your 2 options are conflating the cubemap coord system and how it is viewed. I have my cubemap defined as in the first option. +z forward, +x on the right, y up. However I multiply the skybox coord's z by -1 which places the +z face in the -ve z direction, the direction the default view looks at in OpenGL. If I instead multiply x by -1, then the +z face is in the +z direction but then I'm looking at the -z face.

MarkCallow:
@lexaknyazev please review the latest commit.

lexaknyazev:
Normative and OpenGL parts look good.

In the Vulkan-related section:
> or multiply the Y coordinate by -1

Wouldn't that change handedness and therefore flip faces?

lexaknyazev:
Since rotation by 180 degrees is the same as multiplying two coordinates by -1, the Vulkan options should be:
- multiply both Y and Z by `-1` to align Y and place +Z face in -Z direction, or
- multiply both Y and X by `-1` to align Y and place +Z face in +Z direction.

MarkCallow:
> Wouldn't that change handedness and therefore flip faces?

Yes it does. Crap. How didn't I notice that.

I'll update it and publish again. Sigh!!!

MarkCallow:
Done.

PR #162, opened 2022-01-24
Use dfdTotalSize for DFD definition

lexaknyazev:
Referring to `dfdTotalSize` here makes this part of the KTX spec more readable and better aligned with the upstream DFD spec.

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -166,7 +166,7 @@ UInt32 dfdTotalSize
 continue
     dfDescriptorBlock dfdBlock
           [.optional]#&#xFE19;#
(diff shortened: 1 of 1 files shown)


MarkCallow:
I have no idea what is causing the build failure. I've contacted TravisCI. Given past experience it will likely be a few days before I get a reply.

MarkCallow:
Please rebase against current master and force push. What we think is the fix for the build failure is now in master but on rerunning this PR, the build still fails. The requested actions will get the fix into the branch being used for this PR build.

lexaknyazev:
Done.

PR #173, opened 2022-05-05
Clarify mip-level data layouts

### LINKED ISSUE #152: Mip-level language clarifications
The following issues are relatively minor but still worth fixing.
- [x] The spec does not mention that dimensions of each subsequent mip level are halved compared to the previous one.
- [x] In Section 2, the expressions for `num_blocks_*` do not mention that per-mip dimensions are used there.
- [x] Section 3.13.1 says:
  > When `supercompressionScheme != 0` these bytes are formatted as specified in the scheme documentation.

  However, neither Zlib nor Zstandard scheme notes mention how images are concatenated there. Moreover, the BasisLZ notes technically create a circular reference.

lexaknyazev:
Fixes #152.

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -230,27 +230,27 @@ end
 [[levelImages_defs]]In the `levelImages` loops above,
 
 [stem]
-// num_blocks_z = max(1, ceil(pixelDepth / block_depth))
+// num_blocks_z = max(1, ceil(level_depth / block_depth))
 +++++
-num\_blocks\_z = \max\left(1, \left\lceil{\frac{pixelDepth}{block\_depth}}\right\rceil\right)
-+++++
-
-[stem]
-// num_blocks_y = max(1, ceil(pixelHeight / block_height))
-+++++
-num\_blocks\_y = \max\left(1, \left\lceil{\frac{pixelHeight}{block\_height}}\right\rceil\right)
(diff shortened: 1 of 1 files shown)


ktxspec.adoc
```diff
@@ -671,6 +675,9 @@ transcode target format is selected.
 ====
 
 ===== Zstandard
+* After inflation, the level data follows the uncompressed layout as
+  specified in the <<levelImages>> with <<_mippadding,mipPadding>> omitted.
```
MarkCallow, line 679: Currently when inflating a ktxTexture2, `libktx` inserts the same padding as the data would have had before deflation. There is no mention of this topic in the linked issue though I vaguely recall discussing it somewhere.

What is the reason padding should be omitted? If the image format has multi-byte data then you want it properly aligned for uploading to the GPU.

ktxspec.adoc
```diff
@@ -683,6 +690,8 @@ transcode target format is selected.
   components are present. See <<DFD for Supercompressed Data>>.
 
 ===== ZLIB
+* After inflation, the level data follows the uncompressed layout as
+  specified in the <<levelImages>> with <<_mippadding,mipPadding>> omitted.
```
MarkCallow, line 694: Same comment as above though `libktx` does not currently implement zlib compression.

ktxspec.adoc
lexaknyazev: You're right. I was confused by the following language from the Section 2:
> Applications inflating levels may choose to restore the alignment caused by `mipPadding`.

ktxspec.adoc
lexaknyazev: Done.

MarkCallow:
Thanks.

PR #178, opened 2022-05-06
Update ETC1S slice descriptions

### LINKED ISSUE #158: DFD and BasisLZ grayscale flag
`KHR_DF_CHANNEL_ETC1S_RRR` and `KHR_DF_CHANNEL_ETC1S_GGG` channels are defined as having grayscale encoding but there is no explicit requirement for the global endpoint codebook to be grayscale in this case. Without such a requirement, a global codebook may contain non-grayscale endpoints making the transcoding result undefined.

Two possible solutions:
- Require `isGrayscale` flag to be set when `ETC1S_RRR` is used.
  - This may be difficult to verify on the KTX level because the flag is not byte-aligned and has a variable position in the bitstream.
- Specify slice-internal RGB channels used for transcoding these DFD configurations.
  - Binomial's implementation always uses red slice channel for BC4-5 and EAC targets.

### LINKED ISSUE #159: ETC1S_AAA transcoding
The `KHR_DF_CHANNEL_ETC1S_AAA` description says:
> 1 component: alpha component replicated in all 3 rgb channels.

There are two issues with this statement:
- It cannot be a formal requirement, i.e., receivers of such files should not check that all three channels are the same before transcoding the slice.
- From the encoding side, it may be more efficient to occasionally re-use RGB endpoints that happen to have the same per-channel value (when the channel is fixed).

Binomial's implementation always uses red channel, treating alpha slices as `ETC1S_Axx`.

lexaknyazev:
Fixes #158, fixes #159.

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -926,7 +926,7 @@ the type of data indicated.
 +
 [width=100%,align=center,cols="<30,^10,<60",options=header]
 |===
-| Id | Value | Type 
+| Id | Value | Type
 |`KHR_DF_CHANNEL_UASTC_RGB`  | 0 | 3 component: opaque color. RGB
                                    components in the rgb channels.
 |`KHR_DF_CHANNEL_UASTC_RGBA` | 3 | 4 component: color + alpha. RGB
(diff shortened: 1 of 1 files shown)


ktxspec.adoc
```diff
@@ -959,26 +959,26 @@ The ETC1S image format is indicated by `colorModel` `KDF_DF_MODEL_ETC1S`
...
-|`KHR_DF_CHANNEL_ETC1S_GGG` | 4  | Not used independently. 1 component:
-                                   G component replicated in all 3 rgb
-                                   channels
-|`KHR_DF_CHANNEL_ETC1S_AAA` | 15 | Not used independently. 1 component:
-                                   alpha component replicated in all 3
-                                   rgb channels.
+| Id | Value | Type
+|`KHR_DF_CHANNEL_ETC1S_RGB` | 0  | 3 components: opaque color. RGB
+                                   components in the rgb slice channels.
+|`KHR_DF_CHANNEL_ETC1S_RRR` | 3  | 1 component: R component in the r
```
MarkCallow, line 971: The meaning of "r slice channel" isn't clear.  Do you mean in the slice's r channel? Should we use component instead of channel?

lexaknyazev: > Do you mean in the slice's r channel?

Yes. What would be a better term for it?

MarkCallow: Either "slice's r channel" or "slice's r component" is fine.

lexaknyazev: Switched to using "component" to not confuse with DFD's channels.

PR #180, opened 2022-05-11
Refine mipPadding section

### LINKED ISSUE #175: Confirm mipPadding size
The spec says:
> When `vkFormat` is not `VK_FORMAT_UNDEFINED`, _texel block size_ is as given for the `vkFormat` value in
[section 40.1.6 _Format Compatibility Classes_](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#formats-compatibility) of VULKAN12EXT otherwise it is 16.

- We need to make an exception here for `VK_FORMAT_D16_UNORM_S8_UINT` as the Vulkan table defines it as 3-byte while the KTX spec defines it as 4-byte.
- We need to make an exception here for `VK_FORMAT_D32_SFLOAT_S8_UINT` as the Vulkan table defines it as 5-byte while the KTX spec defines it as 8-byte.
- Does it mean that regular RGB formats like RGB8/RGB16/RGB32/RGB64 need 12 (or 24 for the last case) byte alignment?

lexaknyazev:
Fixes #175.

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -1430,12 +1430,20 @@ bytes are formatted as specified in the scheme documentation.
 `mipPadding` is between 0 and stem:[lcm(texel\_block\_size, 4) -
 1] bytes of value `0x00`. stem:[lcm] is _least common multiple_.
 This is only required when
-`<<_supercompressionscheme,supercompressionScheme>> == 0`. When
-`<<_vkformat,vkFormat>>` is not `VK_FORMAT_UNDEFINED`, _texel block
-size_ is as given for the `vkFormat` value in
+`<<_supercompressionscheme,supercompressionScheme>> == 0`.
+
+_Texel block size_ is as given for the `vkFormat` value in
 {url-khr-vulkan}/specs/1.2-extensions/html/vkspec.html#formats-compatibility[section
-40.1.6 _Format Compatibility Classes_] of <<VULKAN12EXT>> otherwise
-it is 16.
+40.1.6 _Format Compatibility Classes_] of <<VULKAN12EXT>> for all
+`vkFormat` values except the following three:
+
+[options="header"]
+|====
+| VkFormat                     | _Texel Block Size_
+|`VK_FORMAT_UNDEFINED`         | Derived from DFD
+|`VK_FORMAT_D16_UNORM_S8_UINT` | 4
+|`VK_FORMAT_D32_SFLOAT_S8_UINT`| 8
+|====
 
 [NOTE]
 .Padding Rationale


MarkCallow:
I have modified `libktx` to follow these new rules and a bunch of tests are failing. I have to investigate why...

MarkCallow:
> I have modified `libktx` to follow these new rules and a bunch of tests are failing. I have to investigate why...

It was a stupid error on my part. After fixing it, I added some tests for the depth stencil formats and immediately discovered that `libktx` was not giving them the KTX-specified texel block size. I've fixed that. I will create a PR. When it is ready to merge, I'll merge this spec update at the same time.

MarkCallow:
I have realized that `ktxTexture_VkUploadEx` in _libktx_ does not support the depth stencil formats. In looking at implementing support I find there is no easy way to upload these formats to Vulkan. `VkCmdCopyBufferToImage` does not have any way to specify a stride between pixels and the spec. says

> When copying to or from a depth or stencil aspect, the data in buffer memory uses a layout that is a (mostly) tightly packed representation of the depth or stencil data.

Unless I am mistaken or missing some way to specify a stride the only way to upload such data to Vulkan is to have multiple planes. On the other hand the packed `VK_FORMAT_D24_UNORM_S8_UINT` and `VK_FORMAT_D32_SFLOAT_S8_UINT` formats can be uploaded to OpenGL.

Should we continue to support `VK_FORMAT_D16_UNORM_S8_UINT` and `VK_FORMAT_X8_D24_UNORM_PACK32` which can't be uploaded to OpenGL?

Should we make all depth-stencil multiplane thus making the OpenGL loader rather than the Vulkan loader have to do pixel shuffling?

lexaknyazev:
> I find there is no easy way to upload these formats to Vulkan.

AFAIR, we agreed that Vulkan loaders would need to split components manually before upload in that case.

> Should we continue to support `VK_FORMAT_D16_UNORM_S8_UINT` and `VK_FORMAT_X8_D24_UNORM_PACK32` which can't be uploaded to OpenGL?

Overall, I think that KTX should not be too limited by GPU APIs, especially by OpenGL in 2022.

Our format mapping section suggests using `GL_DEPTH_COMPONENT24` for `VK_FORMAT_X8_D24_UNORM_PACK32`, which may be incorrect. We should confirm what happens when an application makes a texture upload with the following args: `internalFormat: GL_DEPTH_COMPONENT24, format: GL_DEPTH_COMPONENT, type: GL_UNSIGNED_INT`. At this point, I think that GL drivers would rescale 32 bits to 24 thus leading to incorrect results.

lexaknyazev:
> Should we make all depth-stencil multiplane thus making the OpenGL loader rather than the Vulkan loader have to do pixel shuffling?

That would technically be a breaking change requiring bumping the major version to 3.

MarkCallow:
> AFAIR, we agreed that Vulkan loaders would need to split components manually before upload in that case.

Yes. The only alternative I see is to create a `VkBufferImageCopy` for each pixel, an even less attractive proposition that splitting before upload.

> Our format mapping section suggests using GL_DEPTH_COMPONENT24 for VK_FORMAT_X8_D24_UNORM_PACK32, which may be incorrect.

I agree with your analysis of what GL drivers will do. Is there an alternative mapping?

> That would technically be a breaking change requiring bumping the major version to 3.

Yes I know. We'll have to stick with what we've got. However I don't think anyone is using these formats yet given the lack of bug reports over the missing functionality in `libktx`.

lexaknyazev:
> Is there an alternative mapping?

No, I'll update the relevant section.

PR #181, opened 2022-05-20
Optimize images

lexaknyazev:
- Recreated the cubemap coordinate system figure and the fixed background element in SVG.
- Optimized the logos to reduce their footprint.
- Removed unreferenced images.

As a result, the built spec is now 65 kB smaller.


docinfo.html
```diff
@@ -44,8 +44,4 @@
   padding-top: 0;
   padding-bottom: 0;
 }
-
-table.dfdexample tr {
-  : 50%;
```
MarkCallow, line 49: I put this in because without it the dfd examples were far too large. Did you do something else to keep them at a reasonable size?

MarkCallow:
We will need a `logoWD` next time we have to make a working draft so either do not delete `logoWD.png` or replace it with an SVG equivalent.

docinfo.html
lexaknyazev: This part doesn't seem to be syntactically correct thus it's a no-op.

lexaknyazev:
> We will need a logoWD next time we have to make a working draft so either do not delete logoWD.png or replace it with an SVG equivalent.

Done.

PR #183, opened 2022-05-22
Use generic fonts

lexaknyazev:
Use `sans-serif` as the main typeface, `monospace` for code listings.
> **Rationale (copied from the glTF 2.0 spec)**: Current CSS file uses `Noto` which may be not a proper typeface name (it should probably have been `"Noto Sans"`). This font is not pre-installed by default, so removing it makes no difference for the most of users. Also, using only the generic font name ensures that the HTML text and the text embedded in SVG look alike, i.e., they both are using `sans-serif` or `monospace` fonts that are default on the user's OS (Arial/Courier on Windows, Helvetica/Menlo on macOS).

### DIFF
--- a/khronos.css
+++ b/khronos.css
@@ -131,7 +131,7 @@ meta.foundation-mq-large { font-family: "only screen and (min-width:1440px)"; wi
 
 html, body { font-size: 100%; }
 
-body { background: #fff; color: #222; padding: 0; margin: 0; font-family: "Helvetica Neue", "Helvetica", Helvetica, Arial, sans-serif; font-weight: normal; font-style: normal; line-height: 1; position: relative; cursor: auto; }
+body { background: #fff; color: #222; padding: 0; margin: 0; font-family: sans-serif; font-weight: normal; font-style: normal; line-height: 1; position: relative; cursor: auto; }
 
 a:hover { cursor: pointer; }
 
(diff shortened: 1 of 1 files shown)


MarkCallow:
`khronos.css` was copied, as was, from the Vulkan spec. Does glTF use the same style sheet as Vulkan or does it have its own. Ideally all should use the same house style.

lexaknyazev:
Ideally they should be split into Khronos-common and WG-specific parts. Since glTF's transition to AsciiDoc happened much later than KTX', glTF includes a newer CSS copied from Vulkan containing even more unrelated styles (like VUIDs).

The typeface issue was fixed ad-hoc in glTF before the v2.0.1 release. Vulkan still has it.

lexaknyazev:
Since KTX and glTF are in the same zone (and both needing a new revision), I suggest we align them first without syncing styles with other specs.

MarkCallow:
> Since KTX and glTF are in the same zone (and both needing a new revision), I suggest we align them first without syncing styles with other specs.

Agreed. Therefore both should use the same stylesheet. After I merge this PR will they be the same?

lexaknyazev:
> After I merge this PR will they be the same?

Sort of. The CSS in the glTF repo defines different colors in some places and has a few unused definitions that came from Vulkan.

PR #188, opened 2022-11-22
Rewrite DFD restrictions section

### LINKED ISSUE #185: Relax DFD transfer function restrictions
The spec says:
> If `vkFormat` is one of the `*_SRGB{,_*}` formats then `transferFunction` must be `KHR_DF_TRANSFER_SRGB` and vice-versa. For all other `vkFormat` values, except `VK_FORMAT_UNDEFINED`, transfer_function must be `KHR_DF_TRANSFER_LINEAR`.

The second rule is impractical for the following reasons:
- 10- and 12-bit data in formats like `A2R10G10B10_UNORM_PACK32` may be legitimately encoded with PQ, HLG, or even sRGB transfer function. The lack of dedicated GPU API formats to achieve correct hardware filtering is a much smaller issue than the KTX 2.0 inability to store the data.
- Legacy packed formats like `R5G6B5_UNORM_PACK16` are almost never used to store true linear values because of the reduced precision. So again, KTX 2.0 cannot be used to store that data correctly.

**Proposal**
Replace the second normative rule with an implementation note explaining common use cases and pitfalls.

lexaknyazev:
- As the section defines restrictions on DFD, the language was updated to assume that `vkFormat` is already set.
- The normative part has been reduced to fix #185.
- The remaining text was rearranged into notes as it does not restrict DFD values per se.

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -871,35 +871,43 @@ A _dfDescriptor_ is useful in the following cases:
 The following restrictions must be obeyed when setting the fields of a
 _dfDescriptorBlock_.
 
-* The value of `<<_vkformat,vkFormat>>` and the image data must
-  never contradict the DFD and, vice-versa, the DFD must not
-  contradict `<<_vkformat,vkFormat>>`. Thus the value of `vkFormat`
-  can be used as a shortcut. Only when it is `VK_FORMAT_UNDEFINED`,
-  does an application need to utilitize sample information from the
-  DFD.  (In this case contradiction is impossible.)
-* `colorModel` must be `KHR_DF_MODEL_RGBSDA`
-  or one of the block compressed color models listed in <<KDF13>>
-  Section 5.6 or its successors, currently `KHR_DF_MODEL_BC1A` to
-  `KHR_DF_MODEL_UASTC`.
-* `colorPrimaries` may be any of the available values since
-  conversion of the selected primaries and white point to a display's
-  can be done simply with a 3x3 matrix multiply.
+* If `<<_vkformat,vkFormat>>` is not `VK_FORMAT_UNDEFINED`, the DFD's
+  `texelBlockDimension*`, `bytesPlane*`, and sample information fields
+  must match the format's definition. The `colorModel` must be
+  `KHR_DF_MODEL_RGBSDA` or the matching block compressed color model
+  listed in <<KDF13>> Section 5.6 or its successors, currently
+  `KHR_DF_MODEL_BC1A` to `KHR_DF_MODEL_UASTC`.
+* If `<<_vkformat,vkFormat>>` is one of the `+*_SRGB{,_*}+` formats,
+  `transferFunction` must be `KHR_DF_TRANSFER_SRGB`.
+* If `<<_vkformat,vkFormat>>` is not one of the `+*_SRGB{,_*}+` formats
+  and an sRGB variant of that format exists, `transferFunction` should not
+  be `KHR_DF_TRANSFER_SRGB`.
 
 [NOTE]
 ====
-`KHR_DF_PRIMARIES_BT709/SRGB` is recommended for standard dynamic range,
-standard gamut images.
+For example, `VK_FORMAT_R8G8B8A8_UNORM` should not be used with
+`KHR_DF_TRANSFER_SRGB` because there is `VK_FORMAT_R8G8B8A8_SRGB`.
+
+On the other hand, `VK_FORMAT_A2B10G10R10_UNORM_PACK32` may be used with
+`KHR_DF_TRANSFER_SRGB` because there is no sRGB variant of this format.
 ====
 
-* `transferFunction` must be `KHR_DF_TRANSFER_LINEAR` or
-  `KHR_DF_TRANSFER_SRGB` since these are the only ones supported by the
-  vast majority of graphics hardware and correct filtering and blending
-  without hardware support is difficult to impossible.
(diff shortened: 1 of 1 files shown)


ktxspec.adoc
```diff
@@ -871,35 +871,43 @@ A _dfDescriptor_ is useful in the following cases:
...
-  does an application need to utilitize sample information from the
-  DFD.  (In this case contradiction is impossible.)
-* `colorModel` must be `KHR_DF_MODEL_RGBSDA`
-  or one of the block compressed color models listed in <<KDF13>>
-  Section 5.6 or its successors, currently `KHR_DF_MODEL_BC1A` to
-  `KHR_DF_MODEL_UASTC`.
-* `colorPrimaries` may be any of the available values since
-  conversion of the selected primaries and white point to a display's
-  can be done simply with a 3x3 matrix multiply.
+* If `<<_vkformat,vkFormat>>` is not `VK_FORMAT_UNDEFINED`, DFD's
```
MarkCallow, line 874: s/DFD's/the DFD's/

ktxspec.adoc
```diff
@@ -871,35 +871,43 @@ A _dfDescriptor_ is useful in the following cases:
...
+  `texelBlockDimension*`, `bytesPlane*`, and sample information fields
+  must match the format's definition. The `colorModel` must be
+  `KHR_DF_MODEL_RGBSDA` or the matching block compressed color model
+  listed in <<KDF13>> Section 5.6 or its successors, currently
+  `KHR_DF_MODEL_BC1A` to `KHR_DF_MODEL_UASTC`.
+* If `<<_vkformat,vkFormat>>` is one of the `+*_SRGB{,_*}+` formats,
+  `transferFunction` must be `KHR_DF_TRANSFER_SRGB`.
+* If `<<_vkformat,vkFormat>>` is not one of the `+*_SRGB{,_*}+` formats
+  and an sRGB variant of that format exists, `transferFunction` should not
```
MarkCallow, line 883: I think s/should not/must not/. Why di d you choose "should"?

ktxspec.adoc
```diff
@@ -871,35 +871,43 @@ A _dfDescriptor_ is useful in the following cases:
...
-  `KHR_DF_TRANSFER_LINEAR`.
+[NOTE]
+====
+`colorPrimaries` may be any of the available values since conversion
+of the selected primaries and white point to a display's can be done
+simply with a 3x3 matrix multiply.
+
+Still, `KHR_DF_PRIMARIES_BT709` / `KHR_DF_PRIMARIES_SRGB` is recommended
+for standard dynamic range, standard gamut images.
+====
```
MarkCallow, line 910: Can we make any recommendation for HDR images? E.g BT.2100.

ktxspec.adoc
lexaknyazev: If future Vulkan versions would have an equivalent of `MTLPixelFormatBGRA10_XR_sRGB`, KTX files created before that with `_UNORM` and sRGB DFD should not suddenly become invalid.

ktxspec.adoc
lexaknyazev: In general, yes. In practice, there are several issues.

- KDFS does not define BT.2100 enum for primaries, although the actual values are the same as in BT.2020, which is defined for DFD.
- 8 bits are not enough to reasonably cover that gamut. All options from 10-bit normalized with PQ encoding to full linear `float32` storage are used in the wild, yet there's no well-accepted standard.
- In some cases, BT.2100 gamut is signalled but not fully utilized (e.g., actual colors stay within P3) for better compatibility.
-  In other cases, DisplayP3 is signalled explicitly and low-precision storage is used, e.g., sRGB curve with 10 bits (or even 8 bits with dithering).

So I don't think there's a simple recommendation that we could put here.

MarkCallow: Thanks. I'll resolve this then and I'll ask Andrew to add BT.2100 to KDFS. It already has a value for the HLG transfer function which is one of the choices in BT.2100.

ktxspec.adoc
lexaknyazev: Done.

ktxspec.adoc
```diff
@@ -871,35 +871,43 @@ A _dfDescriptor_ is useful in the following cases:
...
-  DFD.  (In this case contradiction is impossible.)
-* `colorModel` must be `KHR_DF_MODEL_RGBSDA`
-  or one of the block compressed color models listed in <<KDF13>>
-  Section 5.6 or its successors, currently `KHR_DF_MODEL_BC1A` to
-  `KHR_DF_MODEL_UASTC`.
-* `colorPrimaries` may be any of the available values since
-  conversion of the selected primaries and white point to a display's
-  can be done simply with a 3x3 matrix multiply.
+* If `<<_vkformat,vkFormat>>` is not `VK_FORMAT_UNDEFINED`, the DFD's
+  `texelBlockDimension*`, `bytesPlane*`, and sample information fields
```
MarkCallow, line 875: s/, and/and/. Sorry just noticed this. I haven't used oxford commas anywhere else in the spec., at least not intentionally, so please delete this one.

lexaknyazev: Looks like both styles are used in the text. FWIW, Vulkan style guide requires it.

MarkCallow: Okay. Leave it. I was unaware of the Vulkan style guide and I am surprised to see how many oxford commas there are in the spec. as it is not my writing style. The link from the Vulkan style guide to the Oxford Dictionaries' blog of the video discussion of said commas doesn't work any more.

Issue #194, opened 2023-02-13
Reserve vendor ID for super-compression scheme


JulienVernay-AsoboStudio:
Hello,
I am Julien Vernay, Engine Programmer at Asobo Studio, France.
We would like to use KTX2 in our toolchain, with a proprietary super-compression scheme.
Thus, we need to have a unique ID for the 'supercompressionScheme' field.
Is it possible to have one?
If needed, we have the ASOBO vendor prefix in glTF.

MarkCallow:
> Is it possible to have one?

Yes. You are the first to request one so bear with me for a few days while I get a place set up in the KTX Registry to record the information.

MarkCallow:
@JulienVernay-AsoboStudio please register your scheme via a PR as described in [these instructions](https://github.com/KhronosGroup/KTX-Specification/blob/main/ktxspec.adoc#434-registering-extensions).

MarkCallow:
Done in commit 98043e1eca936938159b5be67e7e5855b8d4883b.

PR #197, opened 2023-03-02
Adding ASOBO proprietary supercompression scheme

JulienVernay-AsoboStudio:
Hello,
This is the patch following up the issue #194 .
I had some trouble with the "icon" causing wrong formatting, so I just dropped it.
Is that correct?


MarkCallow:
There is no "mail" glyph in Font Awesome hence your troubles with the icon. Use `icon:send` as in the example.

For the benefit of the validator it will be beneficial to state if your scheme needs supercompression global data.  Please add either "n/a" or "Required" in the Global Data Format column.

JulienVernay-AsoboStudio:
Thank you for your feedback.
I was misleaded by GitHub viewer, but with AsciiDoctor it works fine :)

MarkCallow:
I realized there was an error in the table header which likely is the reason you have both level data and global data "link" in the same table cell. I fixed the issues and merged this manually. See 98043e1eca936938159b5be67e7e5855b8d4883b.

Thank you for registering your scheme.

JulienVernay-AsoboStudio:
Thank you!

PR #200, opened 2023-08-30
Add VK_KHR_maintenance5 formats

lexaknyazev:


MarkCallow:
Thanks. Are we done? Can I request WG approval for revision 2 now?

vulkan_core.h and related files in KTX-Software and dfdutils need to be updated to pick up these and other recent additions to VkFormat.

lexaknyazev:
Maybe mention the new formats in the Revision History? Other than that, this should be ready.

PR #204, opened 2024-02-06
Disallow 2-plane 444 formats and fix typeSize for _nPACKxx suffix formats.

### LINKED ISSUE #202: Disallow two-plane 444 formats
The [`VK_EXT_ycbcr_2plane_444_formats`](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_ycbcr_2plane_444_formats.html) extension (promoted to Vulkan 1.3) added four new multiplanar formats so the KTX spec should be updated to disallow them.

### LINKED ISSUE #203: typeSize spec for formats with suffix _nPACKxx is wrong
It says typeSize must be n * xx / 8. However it should just be xx / 8 as 'n' here means there are n of the PACKxx quantities and it is the PACKxx quantities will that need swapping.

MarkCallow:
Fixes #202. Fixes #203.

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -2,7 +2,7 @@
 :author: Mark Callow
 :author_org: Edgewise Consulting
 :description: Specification of a container format for GPU textures.
-:docrev: 2
+:docrev: 3
 :ktxver: 2.0
 :revnumber: {ktxver}.{docrev}
 :revdate: {docdate}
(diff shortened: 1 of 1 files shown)


MarkCallow:
@lexaknyazev, @aqnuep please review and let me know if there is anything else we need to fix before submitting to the WG for approval.

ktxspec.adoc
```diff
@@ -2428,6 +2433,10 @@ include::appendices/vendor-metadata.adoc[]
                                       GL formats mapping.
                                     - Increase document width from
                                       55em to 60em.
+|      {docrev}     |  {revdate}  | - Fix typeSize for formats with
+                                      _nPACKxx suffix.
```
lexaknyazev, line 2437: Shouldn't the field name and the format name pattern be wrapped with backticks?

lexaknyazev (approved):

PR #207, opened 2024-02-12
Allow A8B8G8R8 packed formats

### LINKED ISSUE #205: Allow A8B8G8R8 formats
The current rationale says:
> The `A8B8G8R8*PACK32` formats are prohibited because the end result is the same regardless of whether the data is treated as packed into 32-bits or as the equivalent `R8G8B8A8` format, i.e. as an array of 4 bytes, so a Data Format Descriptor cannot distinguish between these cases.

The fact that data layout is the same as with the equivalent `R8G8B8A8` formats does not seem to be a compelling reason to disallow the formats. Tools could be trivially updated to treat them as aliases and all operations would work correctly.

The rationale for supporting them is to ensure that a KTX 2.0 file can be used to capture/replay a Vulkan texture initially created with a `A8B8G8R8*PACK32` format.

lexaknyazev:
Fixes #205.

### DIFF
--- a/formats.json
+++ b/formats.json
@@ -909,6 +909,125 @@
     },
     {
         "type": "PACKED",
+        "vkFormat": "VK_FORMAT_A8B8G8R8_UNORM_PACK32",
+        "typeSize": 4,
+        "blockWidth": 1,
+        "blockHeight": 1,
+        "blockDepth": 1,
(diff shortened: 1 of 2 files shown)


MarkCallow:
Thanks. Please add this change to the change history at the end of the document.

lexaknyazev:
Done.

PR #208, opened 2024-02-19
Update formats.json for compatibility with recent vulkan_core.h and glcorearb.h.

MarkCallow:
Add a compile test for the `generate_format_switches.rb` generated switches.

Prior to the update the switches would not compile when using recent `glcorearb.h` versions due to changes to some tokens from former extensions. They also would not compile using the shortened `vulkan_enum.h` used by KTX-Software due to it omitting extension token names for extensions now in core.

@lexaknyazev please review. Should we create another document revision to publish this or should we just update revision 3? formats.json is the only thing changed.


lexaknyazev (approved):
Feel free to update the revision 3. The formats mapping section is not normative and the changes there are not functional.

MarkCallow:
Thanks @lexaknyazev. I've merged this and force pushed an updated v2.0.3 tag. I'll update the registry once the build completes.

PR #213, opened 2024-06-09
Descriptor Block - incorrect descriptorType bit size

sciecode:
Fixes bit size information on Descriptor Block's [descriptorType](https://github.khronos.org/KTX-Specification/ktxspec.v2.html#:~:text=The%20high%2Dorder%2016%20bits%20of%20its%20first%20UInt32%20are%20the%20descriptor_type).

As can be validated in KDF v1.3 section 4.1 - [descriptorType](https://registry.khronos.org/DataFormat/specs/1.3/dataformat.1.3.html#:~:text=descriptorType%20is%20a%20unique%2015%2Dbit%20identifier), and is also correctly reflected in the example tables above.

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -1247,7 +1247,7 @@ of <<KDF13>>.
 
 ==== dfdBlock
 A `Descriptor Block` as defined in Section 4.1 of <<KDF13>>. The
-high-order 16 bits of its first UInt32 are the `descriptor_type`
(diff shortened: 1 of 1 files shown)


MarkCallow:
Wow! This dates from the very earliest drafts of the spec. and describes KDF v1.2, the version at that time. Well spotted. Thanks for the PR.

PR #215, opened 2024-08-23
Clarify transferFunction restrictions

### LINKED ISSUE #214: DFD transfer function vs vkFormat
@lexaknyazev I'd very much appreciate your input regarding the following.

For the correspondence of the DFD's transfer function with a KTX file's vkFormat field the spec says:

* If <<_vkformat,`vkFormat`>> is one of the `+*_SRGB{,_*}+` formats,
  `transferFunction` must be `KHR_DF_TRANSFER_SRGB`.
* If <<_vkformat,`vkFormat`>> is not one of the `+*_SRGB{,_*}+` formats
  and an sRGB variant of that format exists, `transferFunction` should not
  be `KHR_DF_TRANSFER_SRGB`.

I am modifying `ktx create` to be compliant with the second bullet and I have some questions:

1. Should `VK_FORMAT_ASTC_*_SRGB_BLOCK` be considered a variant of `VK_FORMAT_ASTC_*_SFLOAT_BLOCK`? It is obviously a variant of `VK_FORMAT_ASTC_*_UNORM_BLOCK`.
2. Should we be allowing non-linear transfer functions with unnormalized formats such as `[SU]INT`. If yes should we allow sRGB if there is no sRGB variant?
3. Same question for the SFLOAT formats.
4. Are there any formats for which we should forbid non-linear transfer functions, e.g depth and stencil formats.

I think the answer to Q2 is no to non-linear transfer functions because `[SU]INT` are not normally used for color data.

I think the answer to Q3 is yes to non-linear transfer functions but am unclear on whether to allow sRGB. If sRGB should be allowed then should we not also allow it for `VK_FORMAT_ASTC_*_SFLOAT_BLOCK`?

Not spec-related but relevant to this discussion so I'll ask it here: should we add decode/encode of other non-linear formats to `ktx create`? If so, which? Currently it has ITU (ITU BT.601, BT.709 and BT.2020), ITU BT.2100 PQ, sRGB and gamma decoders and encoders. Note that the command's `--convert-oetf` option only lets you choose `linear` or `srgb`. The ITU and ITU BT.2100 PQ decoders are only used internally to process input files. Should `--convert-oetf` allow more choice of conversions for use with non-sRGB formats? If so, which?

MarkCallow:
after the discussion in issue #214.

Fixes #214.

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -977,6 +977,9 @@ _dfDescriptorBlock_.
     - If <<_vkformat,`vkFormat`>> is not one of the `+*_HLG{,_*}+`
       formats and an HLG variant of that format exists,
       `transferFunction` should not be `KHR_DF_TRANSFER_HLG`.
+* If <<_vkformat,`vkFormat`>> is one of the `+*_[SU]INT{,*}+` formats
+  or one of the depth/stencil formats `transferFunction` must be
+  `KHR_DF_TRANSFER_UNSPECIFIED`.
 
 [NOTE]
 ====
@@ -985,6 +988,12 @@ For example, `VK_FORMAT_R8G8B8A8_UNORM` should not be used with
 
 On the other hand, `VK_FORMAT_A2B10G10R10_UNORM_PACK32` may be used with
 `KHR_DF_TRANSFER_SRGB` because there is no sRGB variant of this format.
+====
+
+[NOTE]
+====
+The `+*ASTC_*_SRGB_BLOCK+` formats are not variants of the
+`+*ASTC_*_SFLOAT_BLOCK+` formats.
 ====
 
 [NOTE]
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -978,8 +978,8 @@ _dfDescriptorBlock_.
       formats and an HLG variant of that format exists,
       `transferFunction` should not be `KHR_DF_TRANSFER_HLG`.
 * If <<_vkformat,`vkFormat`>> is one of the `+*_[SU]INT{,*}+` formats
-  or one of the depth/stencil formats `transferFunction` must be
-  `KHR_DF_TRANSFER_UNSPECIFIED`.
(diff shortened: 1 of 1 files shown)


ktxspec.adoc
```diff
@@ -977,6 +977,9 @@ _dfDescriptorBlock_.
     - If <<_vkformat,`vkFormat`>> is not one of the `+*_HLG{,_*}+`
       formats and an HLG variant of that format exists,
       `transferFunction` should not be `KHR_DF_TRANSFER_HLG`.
+* If <<_vkformat,`vkFormat`>> is one of the `+*_[SU]INT{,*}+` formats
+  or one of the depth/stencil formats `transferFunction` must be
```
lexaknyazev, line 981: To avoid any ambiguities, consider expanding to:
> depth, stencil, or combined depth/stencil

MarkCallow: Thanks for the suggestion.

ktxspec.adoc
```diff
@@ -977,6 +977,9 @@ _dfDescriptorBlock_.
     - If <<_vkformat,`vkFormat`>> is not one of the `+*_HLG{,_*}+`
       formats and an HLG variant of that format exists,
       `transferFunction` should not be `KHR_DF_TRANSFER_HLG`.
+* If <<_vkformat,`vkFormat`>> is one of the `+*_[SU]INT{,*}+` formats
```
lexaknyazev, line 980: Should the pattern be `+*_[SU]INT{,_*}+`?

MarkCallow: Yes. Thanks.

MarkCallow: Should we also say that primaries for these formats must be `KHR_DF_PRIMARIES_UNSPECIFIED`?

lexaknyazev: Yes.

lexaknyazev (approved):

PR #221, opened 2025-01-23
Remove requirement that bytesPlane be 0 for supercompressed.

MarkCallow:
In accordance with upcoming KDF revision.

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -1033,8 +1033,7 @@ uncompressed format before being uploaded to and sampled by a GPU.
 UASTC images can be supercompressed with Zstandard
 (`supercompressionScheme` = 2) with or without first conditioning
 the data with Rate-distortion Optimization. If supercompression is
-used, the DFD's `bytesPlane[0-7]` must be set to 0 as described in
-the next subsection.
+used, the DFD must follow the rules described in the next subsection.
 +
 This color model provides channel Ids, e.g. `KHR_DF_CHANNEL_UASTC_RGB`
 that must be used to indicate the effective number of components
@@ -1123,8 +1122,8 @@ Images in this format are always supercompressed with BasisLZ and
 must be inflated and transcoded to a GPU-supported block-compressed
 format  or decoded to a GPU-supported uncompressed format before
 being uploaded to and sampled by a GPU. Because ETC1S images are
-supercompressed, the DFD's `bytesPlane[0-7]` must be set to 0 as
-described in the next subsection.
+supercompressed, the DFD must follow the rules described in the
+next subsection.
 
 TIP: Whether the image has 1 or 2 slices
 can be determined from the DFD's sample count.
@@ -1148,11 +1147,18 @@ have a sample with a channelType that indicates it is alpha.
 ===== DFD for Supercompressed Data
 When `<<_supercompressionscheme,supercompressionScheme>>` is not 0
 the _dfDescriptorBlock_ must preserve the `colorModel`, `transferFunction`,
-`colorPrimaries`, `flags` and `texelBlockDimension[0-3]` of the
-pre-deflation images along with each sample's `channelType`,
-`qualifiers`, `bitlength`, `bitOffset`, `sampleLower` and `sampleUpper`.
-`bytesPlane[0-7]` must be set to 0 to indicate an _unsized format_,
-as described in Section 5.19 _Unsized Formats_ of <<KDF13>>.
+`colorPrimaries`, `flags`, `texelBlockDimension[0-3]` and
+`bytesPlane[0-7]` of the pre-deflation images along with each
+sample's `channelType`, `qualifiers`, `bitlength`, `bitOffset`,
+`sampleLower`, and `sampleUpper`.
+
+[NOTE]
+====
+Previous document revisions required  `bytesPlane[0-7]` to be set to
+0 to indicate an _unsized format_ in accordance with Data Formats
+v1.3.1 and unpublished drafts of <<KDF13>>. For the foreseeable future
+software must be able to handle such KTX files.
+====
 
 [NOTE]
 ====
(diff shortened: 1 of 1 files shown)


ktxspec.adoc
```diff
@@ -1148,11 +1147,18 @@ have a sample with a channelType that indicates it is alpha.
 ===== DFD for Supercompressed Data
 When `<<_supercompressionscheme,supercompressionScheme>>` is not 0
 the _dfDescriptorBlock_ must preserve the `colorModel`, `transferFunction`,
-`colorPrimaries`, `flags` and `texelBlockDimension[0-3]` of the
-pre-deflation images along with each sample's `channelType`,
-`qualifiers`, `bitlength`, `bitOffset`, `sampleLower` and `sampleUpper`.
-`bytesPlane[0-7]` must be set to 0 to indicate an _unsized format_,
-as described in Section 5.19 _Unsized Formats_ of <<KDF13>>.
+`colorPrimaries`, `flags`, `texelBlockDimension[0-3]` and
+`bytesPlane[0-7]` of the pre-deflation images along with each
```
lexaknyazev, line 1151: Consider adding a note here reiterating that BasisLZ supercompression works only on ETC1S blocks and therefore implicitly requires specific values for dimensions and `bytesPlane*`.

MarkCallow: Addressed in https://github.com/KhronosGroup/KTX-Specification/pull/221/commits/4837e331ba415f9acbe5a90c210a40caaca0ac90. Please review.

ktxspec.adoc
```diff
@@ -1148,11 +1147,14 @@ have a sample with a channelType that indicates it is alpha.
...
+sample's `channelType`, `qualifiers`, `bitlength`, `bitOffset`,
+`sampleLower`, and `sampleUpper`.
+
+BasisLZ supercompression works only on ETC1S blocks so
+`texelBlockDimension[01]` must be 3 and `bytesPlane0` must be
+1 or, if there is an alpha plane, 2.
```
lexaknyazev, line 1157: Pre-deflated length of ETC1S blocks is either 8 or 16.

MarkCallow: Fixed in https://github.com/KhronosGroup/KTX-Specification/pull/221/commits/c5c9f2efa618d7d4135e9e98529ad77b1e230abd.

ktxspec.adoc
```diff
@@ -1148,11 +1147,14 @@ have a sample with a channelType that indicates it is alpha.
...
-`bytesPlane[0-7]` must be set to 0 to indicate an _unsized format_,
-as described in Section 5.19 _Unsized Formats_ of <<KDF13>>.
+`colorPrimaries`, `flags`, `texelBlockDimension[0-3]` and
+`bytesPlane[0-7]` of the pre-deflation images along with each
+sample's `channelType`, `qualifiers`, `bitlength`, `bitOffset`,
+`sampleLower`, and `sampleUpper`.
+
+BasisLZ supercompression works only on ETC1S blocks so
+`texelBlockDimension[01]` must be 3 and `bytesPlane0` must be
+8 and, if there is an alpha plane, `bytesPlane1` must also be 8.
```
lexaknyazev, line 1157: Following the precedence of BC3/BC5 (and also ETC2 RGBA/EAC RG11), only `bytesPlane0` should be used for ETC1S data.

Different DFD "planes" represent disjoint byte sequences, like in YUV 420 formats, that cannot be stored as a continuous stream. Such layouts are not supported in KTX at all.

MarkCallow: > Different DFD "planes" represent disjoint byte sequences, like in YUV 420 formats, that cannot be stored as a continuous stream. 

The example ETC1S DFDs in the current Data Formats draft use bytesPlane[01] so will need to be fixed, if @fluppeteer agrees with your analysis.

lexaknyazev: The issue is a bit more complicated. The KTX spec says the following about ETC1S:
> Images in this format are always supercompressed with BasisLZ ...

_**Tangential question:** This restriction feels inconsistent with the spec design in general. It's one thing to say that a particular supercompression scheme is applicable only to specific payloads. For example, BasisLZ logically operates on ETC1S fields so it's clear why it cannot be used with any other format. But saying that a certain image format must always be compressed with a particular scheme creates a "spec-hardcoded" special case. What if someone proposes an alternative supercompression scheme for ETC1S data? At the very least, I propose changing the current spec language to something like "Images in this format are always supercompressed with a scheme designed for ETC1S streams, such as BasisLZ ..."_

The KDFS spec does not define the pre-deflated state (separate or interleaved) for RGBA or Red-Green ETC1S streams, presumably because they are always wrapped in BasisLZ in practice. One option is to define pre-deflated ETC1S streams as interleaved, i.e., using the same DFD plane like BC3 or BC5; another option is to define pre-deflated ETC1S streams as separate, i.e., using separate DFD planes. The former option would potentially allow using ETC1S with generic supercompression schemes or without any (although it's not very useful), while the latter option would implicitly enforce the requirement of always using a special supercompression scheme with ETC1S.

In any case, KDFS needs to be clarified.

MarkCallow: > The KDFS spec does not define the pre-deflated state (separate or interleaved) for RGBA or Red-Green ETC1S streams, presumably because they are always wrapped in BasisLZ in practice.

It does. See the text beginning at line 3057 in the df.txt draft and the example ETC1S DFDs.

> In any case, KDFS needs to be clarified.

I've opened a thread for this in the MR review on Khronos GitLab. Please add your comment there. You should have received a link.

> > Images in this format are always supercompressed with BasisLZ ...
>
> Tangential question: This restriction feels inconsistent with the spec design in general.
> ...
> "Images in this format are always supercompressed with a scheme designed for ETC1S streams, such as BasisLZ ..."

I don't mind finessing the language. Reading your suggested language made me realize both it and the original are slightly ambiguous. Images could be not supercompressed or, if supercompressed, "done with a scheme designed for ETC1S streams such as BasisLZ." How about "Images in this format are always supercompressed." A scheme designed for ETC1S streams such as BasisLZ must be used."

ktxspec.adoc
```diff
@@ -1148,11 +1148,14 @@ have a sample with a channelType that indicates it is alpha.
...
```
lexaknyazev, line 1158: Nit: this statement does not make it clear that these requirements apply only to BasisLZ. Consider ending it with " when BasisLZ is used" (or rephrasing).

PR #224, opened 2025-02-08
Prepare for approval and release.

MarkCallow:
* Mark as awaiting approval.
* Regularize references to KDFS.
* Verify KDFS reference section numbers.
* State that `KHR_DF_TRANSFER_HLG_UNNORMALIZED_OETF` must not be used.

Actual KDFS reference may need to be updated depending on whether registry file name includes "patch" number.

Speaking of "patch" should I add it to the reference. Currently [KDF13] is used everywhere.

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -2520,6 +2520,8 @@ include::appendices/vendor-metadata.adoc[]
                                       files.
                                     - Add rules for use of transfer
                                       functions.
+                                    - Forbid used of
+                                      `KHR_DF_TRANSFER_HLG_UNNORMALIZED_OETF`.
 |===
 
 [discrete]
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -2,7 +2,7 @@
 :author: Mark Callow
 :author_org: Edgewise Consulting
 :description: Specification of a container format for GPU textures.
-:docrev: 4 (draft)
+:docrev: 4 (for WG approval)
 :ktxver: 2.0
 :revnumber: {ktxver}.{docrev}
 :revdate: {docdate}
@@ -86,6 +86,8 @@ Document Revision 1 approved by the 3D Formats WG Dec 7th, 2022.
 Document Revision 2 approved by the 3D Formats WG Sep 6th, 2023.
 
 Document Revision 3 approved by the 3D Formats WG Feb 14th, 2024.
+
+Document Revision 4 awaiting WG approval.
 
 == Introduction
 
@@ -927,7 +929,7 @@ to read the faces individually rather than as a block of size
 The Data Format Descriptor (_dfDescriptor_) describes the layout
 of the texel blocks in `data`.  The full specification for this is
 is Part 2, Chapters 2 to 11, of the _Khronos Data Format Specification_
-version 1.3 <<KDF13>>.
+<<KDF13>>.
 
 The dfDescriptor is partially expanded in this specification in
 order to provide sufficient information for a KTX file to be parsed
@@ -968,6 +970,8 @@ _dfDescriptorBlock_.
 * If <<_vkformat,`vkFormat`>> is not one of the `+*_SRGB{,_*}+` formats
   and an sRGB variant of that format exists, `transferFunction` should not
   be `KHR_DF_TRANSFER_SRGB`.
+* `transferFunction` must not be `KHR_DF_TRANSFER_HLG_UNNORMALIZED_OETF`.
+  This transfer function is no longer part of the HLG specification.
 * If formats for other transfer functions are added to GPU APIs in the
   future similar restrictions to those just described apply. For
   example, if formats for the HLG transfer function which have the
@@ -1016,7 +1020,7 @@ for standard dynamic range, standard gamut images.
 
 ==== Use of Transfer Functions
 The majority of color spaces listed in Chapter 13 _Transfer functions_
-of <<KDF13>>, with corresponding enumerators given in Chapter 5
+of <<KDF13>>, with corresponding enumerators given in Section 5.8
 _transferFunction_, define only an electro-optical transfer function
 (EOTF) or only an opto-electrical transfer function (OETF) as
 indicated by aliases for the enumerator value. When a KTX file
@@ -1173,7 +1177,7 @@ TIP: Whether the image has 1 or 2 slices
 can be determined from the DFD's sample count.
 
 The bitstream of the ETC1S data pre LZ deflation is described in
-Chapter 21.1 _ETC1S_ of <<KDF13>>.
+Section 21.1 _ETC1S_ of <<KDF13>>.
 
 NOTE: since inflation and transcoding are typically combined in a
 single operation, this bitstream is not visible to applications.
@@ -1306,7 +1310,7 @@ the file is invalid.
 [NOTE]
 ====
 `dfdTotalSize` is included so that the KTX file contains a complete
-descriptor as defined in Chapter 4 "`Khronos Data Format Descriptor`"
+descriptor as defined in Chapter 4 _Khronos Data Format Descriptor_
 of <<KDF13>>.
 ====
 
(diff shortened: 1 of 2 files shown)


ktxspec.adoc
```diff
@@ -1483,12 +1487,13 @@ other combination of these parameters makes the KTX file invalid.
...
+* For compressed color models in Section 5.6 of <<KDF13>> or
+  successors that do not have corresponding Vulkan formats. If the
+  format corresponds to a format in DirectX or Metal then at least
+  one format mapping metadata item is required. One such format
+  exists now, the transcodable format with `colorModel`
+  `KHR_DF_MODEL_UASTC` (= 166). This does not correspond to a DirectX
+  or Metal format.
 * For formats from Metal or DirectX that do not have Vulkan
   equivalents. In this case at least one API's format must be
   recorded in a format metadata item.
```
lexaknyazev, line 1499: I suggest rewriting these points for easier understanding. As currently written, it reads as if UASTC is supported in D3D or Metal until the last sentence of the first point. Then, the second point targets formats supported in APIs other than Vulkan one more time.

Proposal:
> * For any formats from any GPU APIs that do not have Vulkan equivalents.
> * For compressed color models in Section 5.6 of \<<KDF13\>> or successors that do not have corresponding Vulkan formats. One such format exists now, the transcodable format with `colorModel` `KHR_DF_MODEL_UASTC` (= 166).

Then after the list:
> If `VK_FORMAT_UNDEFINED` is used and the format is known to OpenGL, Direct3D or Metal APIs, the corresponding format metadata item should be present.

Note that "must" is changed to "should" here because this requirement is not enforceable in a general case.

ktxspec.adoc
```diff
@@ -2516,6 +2520,8 @@ include::appendices/vendor-metadata.adoc[]
                                       files.
                                     - Add rules for use of transfer
                                       functions.
+                                    - Forbid used of
```
lexaknyazev, line 2523: s/used/use

ktxspec.adoc
MarkCallow: Fixed in https://github.com/KhronosGroup/KTX-Specification/pull/224/commits/a4834ebeafb59cf8902bd1e93305c555d79863b3.

ktxspec.adoc
MarkCallow: Fixed in https://github.com/KhronosGroup/KTX-Specification/pull/224/commits/a4834ebeafb59cf8902bd1e93305c555d79863b3.

lexaknyazev (approved):

MarkCallow:
> Speaking of "patch" should I add it to the reference. Currently [KDF13] is used everywhere.

@lexaknyazev which do you prefer as the reference, [KDF13] or [KDF132]. My preference is towards [KDF13] because it will mean fewer updates to this spec. over time.

ktxspec.adoc
```diff
@@ -926,12 +931,12 @@ to read the faces individually rather than as a block of size
 === Data Format Descriptor
 The Data Format Descriptor (_dfDescriptor_) describes the layout
 of the texel blocks in `data`.  The full specification for this is
-is Part 2, Chapters 2 to 11, of the _Khronos Data Format Specification_
-version 1.3 <<KDF13>>.
+is {url-df-spec}#pass:[_formats_and_texel_access][Part 2, Chapters 2 to
```
lexaknyazev, line 934: What does "Part 2" refer to?

MarkCallow: IIRC once upon a time KDFS was divided into parts with one part for the DFD another part for the color conversion stuff and another for the block compressed formats. It doesn't make much sense now so I'll remove it.

By the way, is there a way to make Asciidoctor check external links when building the doc?

MarkCallow: Fixed in https://github.com/KhronosGroup/KTX-Specification/pull/224/commits/d3e34e57a9154733748effcd088a58116d05cce1.

ktxspec.adoc
```diff
@@ -955,19 +960,22 @@ A _dfDescriptor_ is useful in the following cases:
...
+  currently `KHR_DF_MODEL_BC1A` to `KHR_DF_MODEL_UASTC`.
   `KHR_DF_MODEL_YUVSDA` should be used for all non-prohibited
   `+*_422_*+` formats.
 * If <<_vkformat,`vkFormat`>> is one of the `+*_SRGB{,_*}+` formats,
   `transferFunction` must be `KHR_DF_TRANSFER_SRGB`.
 * If <<_vkformat,`vkFormat`>> is not one of the `+*_SRGB{,_*}+` formats
   and an sRGB variant of that format exists, `transferFunction` should not
   be `KHR_DF_TRANSFER_SRGB`.
+* `transferFunction` must not be `KHR_DF_TRANSFER_HLG_UNNORMALIZED_OETF`.
+  This transfer function is no longer part of the HLG specification.
```
lexaknyazev, line 978: There is no technical reason to disallow this enum. If used (for whatever reason) it would mean that slightly different constants were used during the encoding and the scene-referred values were in the [0:12] range.

Tools may warn that it's officially deprecated though.

MarkCallow: Given that only sRGB and linear were supported in KTX before and that it is not in the current HLG spec, why start allowing it now? It won't break anything. People can use an image processing tool to convert their images when creating KTX files.

MarkCallow: The technical work is already done.

lexaknyazev: It's not about the KTX-Software repo. _Every_ conformant KTX implementation would have to explicitly reject this single enum value although there is nothing really special about it.

FWIW, the KTX spec allows two (!) legacy PAL functions, which are arguably even less useful than unnormalized HLG.

lexaknyazev: For completeness of the argument:

- `KHR_DF_TRANSFER_NTSC` is legacy and normatively ambiguous (see the complete description in KDFS and the list of references)
- `KHR_DF_TRANSFER_PAL_OETF` is also legacy
- `KHR_DF_TRANSFER_PAL625_EOTF` is legacy, deprecated, and commented as "unrealistic"
- `KHR_DF_TRANSFER_ST240` is legacy and deprecated in favour of BT.709

Yet all these transfer functions are allowed.

MarkCallow: Maybe we should prohibit those too! :-) Just joking. I am concerned about viewers having to support all these. I realize one more is not going make much difference to the problem. I think it is okay for viewers to output a note that the file uses an unsupported transfer function so colors may be  off.

lexaknyazev: > I think it is okay for viewers to output a note that the file uses an unsupported transfer function so colors may be off.

Even if a viewer diligently implements all conversions, a user would need a display with the widest possible gamut and enough precision to actually see the correct colors. For instance, NTSC colors cannot be shown on a BT.709 display (although this thread is not about primaries, they are typically used together with the corresponding TFs).

I agree that tools (not the spec) should output some warnings/infos about these TFs.

Ironically, viewing `HLG_UNNORMALIZED` content is no different than viewing regular HLG content as the main difference lies in scene-referred values that are not intended for display anyway.

MarkCallow: I removed the prohibition in https://github.com/KhronosGroup/KTX-Specification/pull/224/commits/c50ec229be1fa463691cb3271822d7a994f54b98. Are you ready to re-approve? I want to merge this and request approval from the WG to publish.

ktxspec.adoc
```diff
@@ -1516,13 +1539,14 @@ hardware vendors and open source developers.
 Formats are identified by one or more of the following:
 
 * A Vulkan VkFormat enum value.
-* A color model value from the Khronos Data Format Specification
-  <<KDF13>>.
+* A color model value from Section
```
lexaknyazev, line 1542: s/Section/Sections

lexaknyazev (approved):
LGTM with couple nits.

ktxspec.adoc
```diff
@@ -1173,7 +1193,7 @@ TIP: Whether the image has 1 or 2 slices
 can be determined from the DFD's sample count.
 
 The bitstream of the ETC1S data pre LZ deflation is described in
-Chapter 21.1 _ETC1S_ of <<KDF13>>.
+Section 21.1 {url-df-spec}#ETC1S[_ETC1S_] of <<KDF14>>.
 
 NOTE: since inflation and transcoding are typically combined in a
```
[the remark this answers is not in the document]
MarkCallow: Fixed in https://github.com/KhronosGroup/KTX-Specification/pull/224/commits/c21466eaa9f2f02cb19c9eec6b5d3eb86d99d285.

ktxspec.adoc
MarkCallow: Fixed in https://github.com/KhronosGroup/KTX-Specification/pull/224/commits/c21466eaa9f2f02cb19c9eec6b5d3eb86d99d285.

PR #227, opened 2025-02-15
Improve math equation typography.

### LINKED ISSUE #226: Multiple formatting issues in STEM expressions
The specification should establish and maintain consistent naming and formatting of symbols and variables used in STEM expressions.

For example, here's the current rendering of some expressions from the "File Structure" section:

Image

There are some issues here:
- All multi-letter symbols are rendered letter-by-letter (check the generated HTML) so kerning for `pixelWidth`, `pixelHeight`, and `pixelDepth` is clearly incorrect.
- Although these three symbols directly refer to the fields defined in the file header, they use the same typeface as the intermediate variables.

Compare with all multi-letter symbols wrapped in `\texttt` (for header fields) and `\textit` (for variables) commands:

Image

There are many more similar examples.

Another issue is inconsistent naming. The spec generally uses `camelCase` for the header fields and `snake_case` for internal variables but `numLevels` and `requiredAlignment` don't follow this.

MarkCallow:
Fixes #226.

### DIFF
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -259,7 +259,7 @@ end
 \textit{num_blocks_x} = \max\left(1, \left\lceil{\frac{\left\lfloor{{\texttt{pixelWidth}}*{2^{-p}}}\right\rfloor}{\textit{block_width}}}\right\rceil\right)
 +++++
 
-where stem:[\textit{p}] is the level index (see <<levelCount>>) and
+where stem:[p] is the level index (see <<levelCount>>) and
 stem:[\textit{block_depth}], stem:[\textit{block_height}] and
 stem:[\textit{block_width}] are stem:[1] for uncompressed formats and
 the block size in that dimension for block compressed formats as given
@@ -852,7 +852,8 @@ An array, `levels`, giving the offset from the start of the file and
 compressed and uncompressed byte sizes of the image data for each
 mip level within the <<_mip_level_array,_Mip Level Array_>> The array is ordered
 starting with stem:[level_{base}] (the level with the largest size images)
-at index _0_. Image for stem:[level_p] will be found at index stem:[p].
+at index stem:[0]. The image for stem:[level_p] will be found at index
+stem:[p].
 
 ===== levels[p].byteOffset
 
@@ -1447,10 +1448,10 @@ When `<<_supercompressionscheme,supercompressionScheme>> != 0` these
 bytes are formatted as specified in the scheme documentation.
 
 ==== mipPadding
-`mipPadding` is between 0 and stem:[\textit{lcm}(\textit{texel_block_size}, 4) -
-1] bytes of value `0x00`. stem:[\textit{lcm}] is _least common multiple_.
-This is only required when
-`<<_supercompressionscheme,supercompressionScheme>> == 0`.
+`mipPadding` is between stem:[0] and
+stem:[\textit{lcm}(\textit{texel_block_size}, 4) - 1] bytes of value
+`0x00`. stem:[\textit{lcm}] is _least common multiple_.  This is
+only required when `<<_supercompressionscheme,supercompressionScheme>> == 0`.
 
 _Texel block size_ is as given for the `vkFormat` value in
 {url-khr-vulkan}/specs/1.2-extensions/html/vkspec.html#formats-compatibility[section
--- a/ktxspec.adoc
+++ b/ktxspec.adoc
@@ -244,32 +244,34 @@ end
 [stem]
 // num_blocks_z = max(1, ceil(level_depth / block_depth))
 +++++
-num\_blocks\_z = \max\left(1, \left\lceil{\frac{\left\lfloor{{pixelDepth}*{2^{-p}}}\right\rfloor}{block\_depth}}\right\rceil\right)
+\textit{num_blocks_z} = \max\left(1, \left\lceil{\frac{\left\lfloor{{\texttt{pixelDepth}}*{2^{-p}}}\right\rfloor}{\textit{block_depth}}}\right\rceil\right)
 +++++
 
 [stem]
 // num_blocks_y = max(1, ceil(level_height / block_height))
 +++++
-num\_blocks\_y = \max\left(1, \left\lceil{\frac{\left\lfloor{{pixelHeight}*{2^{-p}}}\right\rfloor}{block\_height}}\right\rceil\right)
+\textit{num_blocks_y} = \max\left(1, \left\lceil{\frac{\left\lfloor{{\texttt{pixelHeight}}*{2^{-p}}}\right\rfloor}{\textit{block_height}}}\right\rceil\right)
 +++++
 
 [stem]
 // num_blocks_x = max(1, ceil(level_width / block_width))
 +++++
-num\_blocks\_x = \max\left(1, \left\lceil{\frac{\left\lfloor{{pixelWidth}*{2^{-p}}}\right\rfloor}{block\_width}}\right\rceil\right)
+\textit{num_blocks_x} = \max\left(1, \left\lceil{\frac{\left\lfloor{{\texttt{pixelWidth}}*{2^{-p}}}\right\rfloor}{\textit{block_width}}}\right\rceil\right)
 +++++
 
-where _p_ is the level index (see <<levelCount>>) and _block_depth_,
-_block_height_ and _block_width_ are `1` for uncompressed formats and
+where stem:[\textit{p}] is the level index (see <<levelCount>>) and
+stem:[\textit{block_depth}], stem:[\textit{block_height}] and
+stem:[\textit{block_width}] are stem:[1] for uncompressed formats and
 the block size in that dimension for block compressed formats as given
 in the format's section of the Khronos Data Format specification <<KDF14>>.
 
 A _block_ is a single pixel for uncompressed formats and
-stem:[block\_width \times block\_height \times block\_depth]
+stem:[\textit{block_width} \times \textit{block_height} \times \textit{block_depth}]
 pixels for block compressed formats.
 
-For formats whose Vulkan names have `\_422_`, _block_depth_ and _block_height_
-are `1`, and _block_width_ is `2`.
+For formats whose Vulkan names have `\_422_`, stem:[\textit{block_depth}] and
+stem:[\textit{block_height}] are stem:[1], and stem:[\textit{block_width}] is
+stem:[2].
 
 == Field Descriptions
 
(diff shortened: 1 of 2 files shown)


lexaknyazev (changes requested):
Please also include the `KTXwriterScParams` revision history typo fix in this PR.

MarkCallow:
https://github.com/KhronosGroup/KTX-Specification/pull/227/commits/76045a4ea7ae7da8fb038fa463354975ca88be9c addresses most of these review issues. I will close those it addresses.

ktxspec.adoc
```diff
@@ -868,30 +871,30 @@ faces, all rows (or rows of blocks) and all pixels (or blocks) in
 each row for the mip level.
 
 The total size of the image data from
-stem:[levels[numLevels-1\].byteOffset] (i.e., after the first
-`mipPadding`, if any) until the end of the file is:
+stem:[\texttt{levels}[\textit{num_levels}-1\].\texttt{byteOffset}] (i.e.,
```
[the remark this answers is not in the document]
MarkCallow: I'm not sure exactly what you want here. Please elucidate.

ktxspec.adoc
```diff
@@ -868,30 +871,30 @@ faces, all rows (or rows of blocks) and all pixels (or blocks) in
...
 [stem]
 +++++
-levels[0].byteLength +
-\sum_{p=1}^{numLevels-1}
-\left\lceil{\frac{levels[p].byteLength}{requiredAlignment}}\right\rceil
-\times requiredAlignment
+\texttt{levels}[0].\texttt{byteLength} +
+\sum_{p=1}^{\textit{num_levels}-1}
+\left\lceil{\frac{\texttt{levels}[p].\texttt{byteLength}}{\textit{required_alignment}}}\right\rceil
+\times \textit{required_alignment}
```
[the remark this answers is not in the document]
MarkCallow: I'm not sure exactly what you want here. Please elucidate.

ktxspec.adoc
```diff
@@ -1445,10 +1448,10 @@ When `<<_supercompressionscheme,supercompressionScheme>> != 0` these
...
-1] bytes of value `0x00`. stem:[lcm] is _least common multiple_.
-This is only required when
-`<<_supercompressionscheme,supercompressionScheme>> == 0`.
+`mipPadding` is between stem:[0] and
+stem:[\textit{lcm}(\textit{texel_block_size}, 4) - 1] bytes of value
+`0x00`. stem:[\textit{lcm}] is _least common multiple_.  This is
+only required when `<<_supercompressionscheme,supercompressionScheme>> == 0`.
 
 _Texel block size_ is as given for the `vkFormat` value in
 {url-khr-vulkan}/specs/1.2-extensions/html/vkspec.html#formats-compatibility[section
```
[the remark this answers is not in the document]
MarkCallow: If I do that I should update all the Vulkan references to 1.4. Is that okay?

ktxspec.adoc
```diff
@@ -596,22 +600,23 @@ to signal which faces are present.
 Level Array_>> and, by extension, the number of indices in the
 `<<_level_index,Level Index>>` array.  A KTX file does not need to
 contain a complete mipmap pyramid.  Mip level data is ordered from
-the level with the smallest size images, stem:[level_p] to that
-with the largest size images, stem:[level_{base}] where stem:[p =
-levelCount - 1] and stem:[base = 0].  stem:[level_p] must not be
-greater than the maximum possible, stem:[level_{max}], where
+the level with the smallest size images, stem:[\textit{level}_p]
+to that with the largest size images, stem:[\textit{level}_{base}]
```
lexaknyazev, line 604: `base` needs wrapping too.

ktxspec.adoc
```diff
@@ -849,8 +854,9 @@ value must be 0.
 An array, `levels`, giving the offset from the start of the file and
 compressed and uncompressed byte sizes of the image data for each
 mip level within the <<_mip_level_array,_Mip Level Array_>> The array is ordered
-starting with stem:[level_{base}] (the level with the largest size images)
-at index _0_. Image for stem:[level_p] will be found at index _p_.
+starting with stem:[\textit{level}_{base}] (the level with the largest size images)
```
lexaknyazev, line 857: `base` needs wrapping too.

ktxspec.adoc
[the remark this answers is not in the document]
lexaknyazev: We should use at least Vulkan 1.3 because ASTC HDR, forbidden formats from `VK_EXT_ycbcr_2plane_444_formats`, and formats from `VK_EXT_4444_formats` were added to Vulkan 1.3 core and the KTX spec refers to them without `_KHR` or `_EXT` suffixes.

Additionally we could update enums for `VK_KHR_maintenance5` formats because they were added to Vulkan 1.4 core.

All of this should probably be done in a separate PR.

ktxspec.adoc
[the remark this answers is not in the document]
lexaknyazev: `stem:[\texttt{levels[}\textit{num_levels}-1\texttt{\].byteOffset}]`

ktxspec.adoc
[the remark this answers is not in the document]
lexaknyazev:
```
\texttt{levels[0].byteLength} +
\sum_{p=1}^{\textit{num_levels}-1}
\left\lceil{\frac{\texttt{levels[}p\texttt{].byteLength}}{\textit{required_alignment}}}\right\rceil
\times \textit{required_alignment}
```

ktxspec.adoc
```diff
@@ -849,8 +854,9 @@ value must be 0.
...
+at index stem:[0]. The image for stem:[|textit{level}_p] will be found at index
```
lexaknyazev, line 858: Should be `\` instead of `|`.

MarkCallow: Will make a separate PR for the Vulkan references.

I disagree with this change

```
stem:[\texttt{levels[}\textit{num_levels}-1\texttt{\].byteOffset}]
```

and similar. I think the current way with the skinny bracket looks better. The first "levels ..." on the first line in this image is with your suggested change. The second is as it is now.

Screenshot 2025-02-17 at 16 41 10

The other issues are resolved in https://github.com/KhronosGroup/KTX-Specification/pull/227/commits/f37f20c3bdd5a13b8b6dedb75efb29ebfceaafd3.

ktxspec.adoc
MarkCallow: Fixed in https://github.com/KhronosGroup/KTX-Specification/pull/227/commits/f37f20c3bdd5a13b8b6dedb75efb29ebfceaafd3.

ktxspec.adoc
lexaknyazev: > I think the current way with the skinny bracket looks better.

That's fair. My main concern is the extra gap after the dot. How about this:
```
stem:[\texttt{levels}[\textit{num_levels}-1\]\texttt{.byteOffset}]
```

MarkCallow: It doesn't seem to make much difference to the gap after the dot. It makes the spaces on both sides of the dot equal though. Is that what you want? The first expression has the change. The second is the original.

Screenshot 2025-02-17 at 21 42 15

MarkCallow: Regarding the Vulkan references, the registry main page now has only a single link to "latest" which points to a doc with the core spec. + extensions. I think I should remove the version number from the Vulkan references and also change the `vkFormat` section which has separate links to both core and extension. Both 1.2 links are redirected to "latest" by the Vulkan registry.

Off topic for this PR but I want to agree direction before I make a PR to update the links.

MarkCallow: Maybe I should link to https://docs.vulkan.org/spec/latest/chapters/formats.html for the reference in the `vkFormat` section. What do you think? I think the main Vulkan reference should be to the single html spec. document in the registry.

lexaknyazev: Agreed on using the new chapter link for `VkFormat` and the "latest" for everything else.

MarkCallow: I just updated to Asciidoctor 2.0.23. I was previously using 2.0.20. I see very little difference. I even put copies of old and new on separate lines to see if that changed anything.

Screenshot 2025-02-17 at 22 32 48

Your rendering is different in a number of respects.  I wonder why.

I don't want to spend any more time on this so I will make the change.

MarkCallow: Fixed in https://github.com/KhronosGroup/KTX-Specification/pull/227/commits/4f7737a7e4b6b74b35a298772f5ed5dd80fe4817.

ktxspec.adoc
```diff
@@ -863,40 +869,40 @@ for mip level _p_. It is the offset of the first byte after any
...
-`mipPadding`, if any) until the end of the file is:
+stem:[\texttt{levels}[\textit{num_levels}-1\]\texttt{.byteOffset}] (i.e.,
+after the first `mipPadding`, if any) until the end of the file is:
 [stem]
 +++++
-levels[0].byteLength +
-\sum_{p=1}^{numLevels-1}
-\left\lceil{\frac{levels[p].byteLength}{requiredAlignment}}\right\rceil
-\times requiredAlignment
+\texttt{levels}[0].\texttt{byteLength} +
```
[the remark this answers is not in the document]
MarkCallow: Fixed in https://github.com/KhronosGroup/KTX-Specification/pull/227/commits/f5f61b85968c7739adb0c8081b14997702043b84.

lexaknyazev (approved):
I think this PR is good to go but there is a new issue.

Apparently, all `\textit` blocks are rendered incorrectly (using the Times font) with the default "HTML-CSS" MathJax renderer in Chrome. This issue is not present in other browsers. Switching the renderer to "Common HTML" (right-click on any math expression) also bypasses the issue.

I'll take a look.

lexaknyazev:
The "HTML-CSS" MathJax renderer replaces regular letters used outside of `\textit` blocks with special Unicode math characters. For example, `p` (0x70) is replaced with `𝑝` (0x1D45D). Then all such symbols (regular and special) use the `font-family: STIXGeneral-Italic` style in Chrome and Safari.

For some reason, Chrome (at least on macOS) does not recognize the `STIXGeneral-Italic` font family and the following happens:
- Special math symbols fall back to `STIX Two Math` and the issue remains hidden.
- Regular symbols fall back to `Times` (or whatever the default font face is) and the issue appears.

Safari, on the other hand, does recognize `STIXGeneral-Italic` font family and internally replaces it with `Times-Italic` for regular symbols thus hiding the issue completely.

The proper solution is to switch to the "CommonHTML" MathJax renderer, which is platform-independent. The simplest way to do that is to post-process the generated HTML and replace `TeX-MML-AM_HTMLorMML` with `TeX-MML-AM_CHTML`. The existing `inline-images.rb` script could be updated to perform this change like this:

```diff
diff --git a/inline-images.rb b/inline-images.rb
index cb54f67..130bd0e 100644
--- a/inline-images.rb
+++ b/inline-images.rb
@@ -6,6 +6,7 @@ require 'base64'
 Asciidoctor::Extensions.register do
   postprocessor do
     process do |document, output|
+      output.sub!('TeX-MML-AM_HTMLorMML', 'TeX-MML-AM_CHTML') if document.basebackend? 'html'
       _replace_urls(output) if document.basebackend? 'html'
     end
   end
```

MarkCallow:
Thanks for the investigation into the MathJax issue. I've added your suggested workaround. Please file a bug with Chrome.

lexaknyazev:
This seems to be a macOS "feature" rather than a Chrome bug. The "STIXGeneral" font was supported up until macOS 12 but then it got demoted to "Document-support" starting with macOS 13. In the latter case, fonts seem to become less accessible to applications. For reference:
- https://support.apple.com/en-us/103203
- https://support.apple.com/en-us/103197

MarkCallow:
> This seems to be a macOS "feature" rather than a Chrome bug. The "STIXGeneral" font was supported up until macOS 12 but then it got demoted to "Document-support" starting with macOS 13. In the latter case, fonts seem to become less accessible to applications.

What struck me as a bug and led to my comment was replacing a missing italic font with a non-italic font when "STIXGeneral" is not recognized. Did I misunderstand your original report?

Unrelated, is there a way in Asciidoctor to tell it to use "CommonHTML" in the Mathjax it generates?

lexaknyazev:
> replacing a missing italic font with a non-italic font when "STIXGeneral" is not recognized

That behavior does not seem to be specified anywhere so browsers do as they wish. I think the issue is mainly caused by directly using `STIXGeneral-Italic` as a font name. If it was just `STIXGeneral` with `font-style: italic`, the fallback formatting would work much better.

> Unrelated, is there a way in Asciidoctor to tell it to use "CommonHTML" in the Mathjax it generates?

No, it's hard-coded. Also the MathJax version bundled with Asciidoctor is quite old.

PR #228, opened 2025-02-18
Update vulkan references

MarkCallow:


ktxspec.adoc
lexaknyazev, line 1: Should the reference on L339 be updated?

MarkCallow: Thanks. Fixed in https://github.com/KhronosGroup/KTX-Specification/pull/228/commits/00bc98fe720a8f920f58b59917ca61477511f896.

PR #230, opened 2026-03-10
Create CONTRIBUTING.md

KhronosWebservices:
Add a basic Khronos CONTRIBUTING file to help both Khronos members and non-members understand the Contributor License Agreement (CLA).


MarkCallow (commented):
Thanks for this James. However it appears to be for contributions to an open source repo. This is primarily a specification repo, though it does have a single Ruby script that is for use by external projects rather than being used to generate the specification. Doesn't this file need to cover both types of contribution?

PR #233, opened 2026-06-23
Fix workflow

MarkCallow:
Make it work in the face of Node updates and Cloudflare challenges.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@v7
 
     - name: Install Asciidoctor
       run: sudo apt-get install -y asciidoctor
@@ -49,7 +49,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@v7
 
     - name: Generate switches
       run: make switches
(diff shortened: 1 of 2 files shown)


xavier2k6:
@MarkCallow 

* `actions/checkout` version can be updated to `v7` -> https://github.com/actions/checkout/releases
* `actions/deploy-pages` version can be updated to `v5` -> https://github.com/actions/deploy-pages/releases

~Where's the `cloudflare` issue being shown?~ I see it.

MarkCallow:
Thanks @xavier2k6. I have updated the actions as you suggested.

MarkCallow:
@xavier2k6, any ideas on how to fix the cloudflare issue here? I do not understand exactly what part of the change you made over in install-nsis fixed the cloudflare issue. I mimicked that as closely as I could when using curl directly.

MarkCallow:
@xavier2k6, Pinging you again. If you can't offer any guidance for solving the cloudflare, please at least explain how the changes you made in `install-nsis` fixed the issue there.

MarkCallow:
> @xavier2k6, Pinging you again. If you can't offer any guidance for solving the cloudflare, please at least explain how the changes you made in `install-nsis` fixed the issue there.

What I have now learned is that the user agent you specify in a wget or curl command has to be registered with the site from which you are downloading. I have fixed the problem here by getting the download site to register with Cloudflare a particular user agent string which this CI now passes. Once this CI got past Cloudflare, it tripped up over GH complaining "too many requests". To fix that I have to pass the GITHUB_TOKEN to GH by setting `--header="Authorization: ..."` on the wget command lines.

I suspect that adding a user-agent to the `curl` command in `install-nsis` is what fixed the issue there but only because the specified user-agent had been registered by Source Forge. Knowing this detail a week ago would have saved me a lot of time.

PR #235, opened 2026-07-24
feat: add DXGI-to-VkFormat switch generation script

walcht:
Some consumers may need a specification-compliant (and much easier to integrate) auto-generated DXGI-to-VkFormat switch. E.g., DDS to KTX2 converters have to figure out the corresponding VkFormat of the input DDS DXGI format.

Related: https://github.com/KhronosGroup/KTX-Software/pull/1223


MarkCallow:
Looks good. Would be great to extend `switch_test` to cover this. I know at present it only covers the VkFormal to GL switches. I am planning to  add vk2dxgi and vk2metal tests when I add those queries I mentioned to KTX-Software. For those queries I have created header files with the DXGI and Metal formats which can be used for those tests.

walcht:
> Would be great to extend switch_test to cover this.

Done. By the way, generating DXGI->VkFormat mappings from reverse VkFormat->DXGI might not be ideal because multiple DXGI formats can (need to verify) map to the same VkFormat.

E.g., see `DXGI_\w+_TYPELESS` formats. These are currently not supported but can (as far as I understand) map to some VkFormats. Same with DXGI formats with the `X\d+` placeholder.

At the very least, we can be confident that the generated switches, even though they cover only a subset, are 100% correct and with no ambiguity.

MarkCallow (approved):

MarkCallow:
> By the way, generating DXGI->VkFormat mappings from reverse VkFormat->DXGI might not be ideal because multiple DXGI formats can (need to verify) map to the same VkFormat.
> 
> E.g., see `DXGI_\w+_TYPELESS` formats. These are currently not supported but can (as far as I understand) map to some VkFormats. Same with DXGI formats with the `X\d+` placeholder.

We must look into this but keep in mind that
* some VkFormats are not allowed in KTX, mostly multi-plane YUV formats and formats used for vertices rather than images
* and some VkFormats with the X placeholder do not exist outside a Vulkan implementation. That is they cannot be uploaded to the GPU. That is why KTX defines specific mappings for depth and stencil formats which have X placeholders.

